PR #6, opened 2019-12-07
Map Stuff

Kazkin:
The basic map skeleton for all the map contributors to see. On the map are also 3 tomato mobs named Floor 1 - Surface, Floor - 2 Underground, and Floor 3 - Underground that you can find via the ghost command follow, this is so the general community players can look at the map while observing in an easy fashion. These will be removed layer. Also included is a skeleton of the colonies hallways as a semi-guide.

This PR was far more work than needed but fuck it I signed up for this.


Kazkin:
Added my mech re-balance PR to this as well since it wasn't merged yet.

skraaak:
Good to go. Kinda was hoping this would have been speedmerged by now. Not like we're all waiting to work on the map or anything.

PR #79, opened 2020-04-28
Adds a regulation haircut

DerFlammenwerfer:
Useful for current and former Solfed troops.

Also adds the standard gripper to all(?) robot modules.


LordPapalus:
I'm going to assume this is tested and working

PR #122, opened 2020-06-01
This is going to break ak47s

Kazkin:
But I don't five a single fuck, will fix it after this PR is merged.


Kazkin:
Ignoring the Travis because I still don't give a single fuck.

PR #167, opened 2020-06-23
Does this fix the crashes? - Disables notes in stashes

PolymorphBlue:
Revert this when either BYOND is fixed or the crashing keeps happening.


PolymorphBlue:
Let's not merge this yet, then, after speaking with Papulus.  The notes are the only way to find the stashes, apparently, and according to him the issue is dependent on playercount too?

PR #324, opened 2020-08-24
Bugfix Update

Kazkin:
Fixes mines so they can be better used in traps and allow for more functional use instead of just "X is a hazard". Croaker lords properly open their eyes when they begin search for a target. Armor now doesn't globally update when upgraded, fixing the bug we had.


Kazkin:
I don't really give a fuck what the Travis has an issue with. The code works so fuck off.

PR #428, opened 2020-09-28
Added a Kaiser in the back of the roach cave.

KathyRyals:
Title.

image


Kazkin:
I like this idea but do not want an absolute spawn, in my map changes I'll add a 50/50 random spawner in the roach cave.

PR #454, opened 2020-10-02
Added rice hat.

KathyRyals:
Added a rice hat, with the sprites provided by Zienite.


Kazkin:
I'll merge this for now but we really should make a crafting recipe for rice to make this a hat using dried rice.

PR #487, opened 2020-10-14
Renames Warrant officer to Commissioner

Lamasmaster:
*Renames Warrant Officer (military rank) to Marshal Commissioner (Actual title used for police chiefs)
*Gives them actual prefix that would show up in chat
*(if approved also renamed loadout and character selection text)


Kazkin:
A neat idea but I'm not sold on the name commissioner, warrant officer is fitting giving how tied to the blackshield the marshals are.

PR #505, opened 2020-10-23
Map Adjustments.

Kazkin:
Adds the bullet fab, fixes some visual displays, makes disposals less likely to jam on the surface floors, and changes all map transitions teleporters to be dynamic so adding in additional floors no longer causes them to be fucked.


Kazkin:
Waving the 24h wait on this due to it being minor map edits that, if broken, the travis would have screamed. All changes were also tested.

PR #597, opened 2020-11-25
Mech armor penetration rework

Hydroxyapatite:
## About The Pull Request
<summary>

* Reworks the mech armor penetration calculation to fix a bug where penetrating rounds would knock pilots out of their mechs while not properly ejecting them. This bug was caused by existing code allowing penetrating rounds to run their on-hit proc against the pilot directly. This rework fixes this bug by removing this mechanic and replacing it with the one detailed below.

* Weapons with penetration will now deal additional damage to mechs with lighter armor.  For example, an AMR will now take down a Ripley in two shots. The intent here is to make heavier weapons more desirable to use against mechs. Depending on how this plays out, I may try making small arms deflect more often against heavily armored mechs, say, if the difference in penetration value versus mech armor is two or more.

* A weapon's penetrating value will now affect how likely it is for a component in a mech to be damaged. Added a message for when this happens. I'm not sure this internal damage actually does anything, though. May be a future fix.

* Also removes the power drain from rubber bullets because why is that a thing.
</summary>

	
</details>

## Changelog
:cl:
balance: penetrating weapons will now deal additional damage to mechs depending on how much their penetrating value beats the mech's armor.
fix: fixed a bug where mech pilots would be knocked out of their mechs when shot.
/:cl:


Kazkin:
Nice and clean code wise, merged.

PR #629, opened 2020-12-02
Defib hotfix

Kazkin:
Derp.


Kazkin:
Hydra did a better version, closing this and preserving the lazy hack method in case it proves useful later.

PR #683, opened 2020-12-17
Disables Cthmant Augmentations

Risikio:
No longer allows Cthmants to alter their bodies in Character Setup.

Disables the Cruciform "Background" for Cthmants in Character Setup.

I do not know what will happen to any Cthmants with augmentations already saved to their loadouts.

I do not know what will happen if someone decides to try and install a Cruciform (or metal appendage) during the course of gameplay.


Kazkin:
I'll use the code here to modify a more refined version, I'll close this but reference it later.

PR #814, opened 2021-02-02
Render Patriarch

MSieghardt:
**Render Patriarch**

- A massive red beast of 2200 health.
- Deals an initial damage of 30 - 35 and runs slightly faster than a normal person.
- Tires itself out with every melee attack, lowering damage and speed. 
- It has three abilities

1. A charge which deals a small amount of damage, will throw people aside and can be cast with Middle Mouse Button, low cooldown.
2. A howl that will stun anyone without ear protection, much like a flashbang, long cooldown.
3. A spine burst that shoots low damage spines in a spiral around it, medium cooldown.

- Using an ability will remove fatigue from the Patriarch and heal him for a very small amount (roughly 1% of its health)
- It can change its view distance from 7 (default) to 9 to 12.
- It drops by default a kit coresponding to a random department. Seeing as it's an admin spawn, it should however have other rewards as well spawned by said admin.
- **Should always be controlled by a player right now. By default it will be a very weak creature that will tire itself out and not use abilities. AI to be added in the future.**
- All abilities are telegraphed by both text and animation.

Tested and runs without causing issues. Balance not tested (I don't have with who!)


code/modules/mob/living/simple_animal/hostile/huntersprey.dm
```diff
@@ -0,0 +1,338 @@
...
+/mob/living/simple_animal/hostile/renderpatriarch/UnarmedAttack()
+	..()
+	if(fatigue <= maxfatigue)
+		fatigue = fatigue + 1
+	if(fatigue > 0 )
+		speed = initial(speed) * fatigue
+		melee_damage_lower = initial(melee_damage_lower) - fatigue * 2
+		melee_damage_upper = initial(melee_damage_upper) - fatigue * 2
+
+/mob/living/simple_animal/hostile/renderpatriarch/death(gibbed, deathmessage = "lets out a gutteral snarl before it crumbles to the ground dead!")
```
Muffindrake, line 82: Guttural.

MSieghardt: https://github.com/sojourn-13/sojourn-station/blob/74e60b1baf93e3e3ead7c7b2c5f9c8381895f9e7/code/modules/mob/living/simple_animal/hostile/big.dm#L65

That's one of the lines I **didn't** change. Haha! I'll take care of it.

PR #951, opened 2021-03-07
Balance (stats) change for (some) weapons. 

Chevalt:
About The Pull Request : 
(Now fixed.)
This is the change that was announced after modifying 257. and 408. stats, after an awful fight with my poor coding skills, I've managed to put all of those changes in a single PR. However, the fact it can't merged together automatically might be because some projectile values are the same. 
(Despite being a change from the forked version of sojourn, which was somehow different from the version the server uses on SS13.)
Some weapons have also been modified, like the Scout and SA AKM to actually be relevant now.
🆑
Weapon balance.
/🆑


Kazkin:
Sadly this has conflicts.

Trilbyspaceclone (changes requested):
- Do NOT up recoil
- Scout heavy bolt gun is fine as well as other gun damage tweak

PR #959, opened 2021-03-11
Guild parts rebalance

ThePainkiller:
## About The Pull Request

- Lowers manipulator plasteel cost, it was as costly as crafting the blackbox wich is an anomaly
- Changes borosilicate for reinforced glass, as borosilicate is incredibly rare to find in considerable quantities
- Corrected light overlay positioning for the hard hat since it was **way** off

**EDIT:** Platinum might be rarer and thus could be justified to be part of the Capacitor crafting, but gold and silver are higher conductive materials and thus make more sense, I've decided to leave the recipe unchanged thus.

## Changelog
:cl:DimmaDunk
tweak: Guild matter bins take reinforced glass to make instead of borosilicate.
balance: Guild manipulators no longer take an ungodly ammount of plasteel to craft.
/:cl:


Kazkin:
After some consideration, I'd prefer to keep the borosilicate requirement, as tier 4 parts are supposed to be quite difficult to obtain, the borosilicate requirement necessitating either working with mining, science, or prospectors depending on availability.

ThePainkiller:
> After some consideration, I'd prefer to keep the borosilicate requirement, as tier 4 parts are supposed to be quite difficult to obtain, the borosilicate requirement necessitating either working with mining, science, or prospectors depending on availability.

Fair enough, if it's only one sheet it's kinda affordable, but even with mining active I have rarely ever seen more than 14 sheets of borosilicate together, from personal experience
~~Not like anyone even uses matter bins outside of protolathe anyways~~

PR #961, opened 2021-03-11
Cheaper kinky costumes

ThePainkiller:
## About The Pull Request

Cigs got cheaper but the kinkmate still has really inflated prices, this makes them more affordable in range (considering it is now harder to get on the new map, this might encourage more use out of it)

## Changelog
:cl:DimmaDunk
balance: Kinkmate now has cheaper costumes, for all your degenerate needs.
/:cl:


Kazkin:
Okay pervert

PR #467, opened 2020-10-05
Makes the subtle verb hidden to ghosts

wb13:
## About The Pull Request
<details>
<summary>
Conceals the content of the subtle verb from ghosts - for when players really only want those close by to see.
Logs are still created as appropriate.
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

PR #1002, opened 2021-03-25
Revert "Makes the subtle verb hidden to ghosts"

king5327:
Reverts sojourn-13/sojourn-station#467
I'm annoyed that this slipped my watch, after all those times I'd explicitly rejected this idea.
Also didn't appear on the changelog.

Caught it by accident while trying to figure out why whispers weren't being logged, of all things.


ThePainkiller:
Wasn't this basically intended for people who didn't want their (impure) actions literally be seen by ghosts? The description even said so. The vars were defined for it to work this way. Why do you want to revert this, if it was intended all along?

king5327:
Intended... by whom? Subtle was for short range emoting like whispers, not for eroticism.
t. guy who ported subtle years ago

Trilbyspaceclone (changes requested):
This is a bad change, anti-subtle verson is for ERP without annoying ghosts

king5327:
I can assure you that if I wanted subtle to be antighost, I would have done that when I first ported it.

Kazkin:
Request from the host himself, k5 rarely asks for things and often allows alot of things he wants to pass by unused/undone. I'm more than willing to accept this as its a minor change.

PR #1089, opened 2021-04-16
[FIX&REBALANCE]Revival Sickness

Yawet330:
## About The Pull Request
<details>
<summary>
	Fixes Revival Sickness to be properly scaled with the severity of it.
</summary>
<hr>
This was a rather strange thing to come across - Revival sickness used to be extremely strange on how it worked, going from a -15 debuff, to a -10, to a -25 in terms of damage modifiers. This corrects that, and the stat-loss.

While this actually does nothing in terms of damage incoming, it now further reduces skills.

	
<hr>
</details>

## Changelog
:cl:
Tweak: Revival sickness now scales with its tier, along with applying lesser/much worse stat debuffs
/:cl:

### DIFF
--- a/code/datums/perks/job.dm
+++ b/code/datums/perks/job.dm
@@ -265,22 +265,22 @@
 	..()
 	initial_time = world.time
 	cooldown_time = world.time + 30 MINUTES
-	holder.brute_mod_perk += 0.15
-	holder.burn_mod_perk += 0.15
-	holder.oxy_mod_perk += 0.15
-	holder.toxin_mod_perk += 0.15
-	holder.stats.changeStat(STAT_ROB, -15)
-	holder.stats.changeStat(STAT_TGH, -15)
-	holder.stats.changeStat(STAT_VIG, -15)
+	holder.brute_mod_perk += 0.10
+	holder.burn_mod_perk += 0.10
+	holder.oxy_mod_perk += 0.10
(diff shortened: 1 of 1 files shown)


Kazkin:
Not sure on this balance change. going to let it sit for a bit and see what others think.

ThePainkiller:
The brute loss is okay, but **TWENTY** statloss across the board if you have less than 40 Toughness? That's a bit too severe...
Consider the average colonist rarely, if ever, reaches more than thirty unless lucky anomalies and a lot of leveling, plus the ones with higher toughness tend to be part of hard risk jobs like Prospectors and Blackshield.
Sadly this still won't enforce people buckling themselves to a bed to receive **mere two less seconds of the debuff**, and doesn't still fix the real issue:
Severe Revival Sickness reduces merely MEC+COG+BIO, effectively being a worthless punishment for anyone whose Toughness is between 40 and 60, making you never want to go **above** these values, enforcing stat minmaxing.
In my opinion a good middle ground would be:
- Reduce damage penalties on severe to `0.2` instead of `0.25` since it accomodates the scaling you're giving it
- Add ROB+TGH+VIG penalties to Severe Revival Sickness so that it makes sense (why would someone with above 40 toughness be penalized by having reduced mental stats, but if you have more than 60 you get actually penalized on combat stats?)

Kazkin:
After some consideration, merged.

PR #1111, opened 2021-04-19
Robotics plasteel balance

AysheDaArtist:
Ripley now requires use of: Heavy Cell Mount, Hydraulic Circuits, and high power motor which equals 10 Plasteel in total to build.  This makes it so Robotics can at least make a Ripley or use it elsewhere and make Miners mad.


Kazkin:
Conflicts.

AysheDaArtist:
Sorry!  Still working on how to do PRs and not mess stuff up.
All I did was go inside mapping and change the value to the plasteel inside it to be 20.
Sorry if this was the incorrect way to change that value.

PR #1148, opened 2021-04-28
Possibly Improves Character Setup Lag

Risikio:
Comments out commands to create a east looking preview model that is never used. 

Should improve performance issues regarding mannequins, possibly by 25%?


king5327:
Ok so, uh.
The one image in the main character creator, it makes sense to remove.
However, when you're editing augmentations, it lets you rotate the character in all directions.
So it doesn't make sense to disable that.

king5327:
Looking at it again, yeah, our lag isn't so much rooted in update_icon as it is in everything that happens before. Sure, this might help a little, but it addresses something later in the sequence than what I believe is the issue.

PR #1196, opened 2021-05-05
Categorizes Graphic Updates in Char Creation

Risikio:
Now instead of update_preview_icon sanitizing the entirety of the character every time, only the parts of the mob that belongs to the altered category is changed.

I am now officially out of ideas of how to better optimize the processes of creating a character.


king5327:
I'm going to mention that this probably breaks character creation entirely.

PR #1225, opened 2021-05-09
Time Tracking

MSieghardt:
This makes it so that we track play time per faction on the server for each client.

Admins can add time, both through a command and by editing the variables. 

Time checks tick every 5 minutes.

### **Please review this change in depth before merging, this affects the save files for preferences. It has been tested and everything works as desired / expected, but this was at a small scale, up to 30 minutes ingame time testing. Please make sure K5 approves of this before merging.**


king5327:
My main complaint with this comes from the departments being hardcoded in *yet another* part of the code.

MSieghardt:
@king5327 I tried to use the defines where I could, such as in the 'if' clauses. I agree it would be best if we could make it all modular, but when I was doing some tests with a list that used the defines or some way that would account for new departments, I ran into compiling issues initially and after jury-rigging it, some save-file issues.

At the moment the main place where we don't have the defines used is in the client.prefs place, with each department having its own variable there, with them being called other places. Got any idea how to work into implementing this with a list or some way that would use those defines?

Kazkin:
This one is up to k5 to merge.

king5327:
Writing this at work, but ideally, we'd have the department datums defining how they're checked and saved. Here we'd just be feeding the players to each datum. Something like that.

MSieghardt:
@king5327 Updated to be a lot more modular. Proc is in job datum now.

Kazkin:
LJHVDSRHODVHOLIVDRHOIBRT

PR #1268, opened 2021-05-19
Small Lemniscate Buff + Description Fix

intense-skies:
As it says on the title.
-The buff itself is a +5 to the previous amount by which stats are increased. Moving it from +10 to +15.
The reason behind this is that Lemniscate is a bit underwhelming as a path. Simply having longer duration doesn't give them much at all. You could do the same thing lemniscate's do as a vector especially since you'll likely be sticking around the people you buff, such as during events. Lemniscates are now more powerful within their niche.
-The second thing adjusted is the description of the litanies so that it does its job correctly. Previously they were bugged with calling the path to the short_boost version.


intense-skies:
Seems like my PR is pulling in extra stuff. I'm shit with github so its going to be closed for now.

PR #1282, opened 2021-05-22
Church flare litany fix + nerf to it

intense-skies:
What this PR does:
-Fixes a broken litany which was commented out.
-Slight description adjustment to give an idea for its duration.
-Also nerfs the litany with: 
   --Cost made greater with the original of 15 to new 30. 
   --And light level produced from 6 to 5. A lantern's light level is 4.


ThePainkiller:
I can't believe it....the first church nerf in history....

intense-skies:
Not the first nor will it be the last.

intense-skies:
Adjustment does: 
-Makes it brighter back to the original 6
-Adds 2 minute cool down (can be bumped up).
-Removes useless line of code

This is because of consideration to the glowing book litany.

The purpose I see behind the flare litany is a stationary great light source while the book litany is a portable good light source. 2 minute cd means that vectors/primes can't spam it but can still eventually light up a place as needed.

PR #1285, opened 2021-05-22
New Church Litany: Word of Guidance aka Glow Book + Ritual book storage adjustment

intense-skies:
What this PR adds:
1. A base litany that any cruciform bearer can access.
2. Allows the church's ritual book fit within the belt storage slot
This litany does the following:
-Makes the ritual book the individual is holding glow a rather bright (brightness 5) level of light. 
-Costs 15 energy, lasts 15 minutes, has a 15 minute cool down, and is one level brighter compared to a lantern.
It also plays a nice holy sound when you perform it plus letting everyone else around you including yourself see that your book starts glowing.
Balance thoughts: The closest and easiest comparison is to a lantern. A lantern is easy to get, is slightly dimmer, but it fits into your pocket or the back slot of something you're wearing. The ritual book only fits into your belt slot, where a common tool belt or webbing would go, or has to be held in your hand. So I think its rather decent in general.
In general I think its a nice church-flavored addition which provides something cool.  If there is any thoughts/ideas/concerns let me know. It has been tested nearly entirely except for the AoE message. But I'm pretty certain that works as its mostly copied code. Other then that small thing everything else has been checked.


Kazkin:
Conflicts.

intense-skies:
Going to make a new PR since I'm not good at fixing branches.

PR #1305, opened 2021-05-24
Comment out assisted organs.

r4d6:
## About The Pull Request
Assisted organs are dumb. This remove Assisted Organs

They are repaired like prosthetic organs.
BUT
They aren't affected by EMP like prosthetic organs are.
AND
They get healed/hurt by chems like organic organs.

Not to mention that they aren't mentionned ANYWHERE in the code. Even the Prosthetic organs had 1-2 actual organs coded in.

If you have assisted organs on your characters, fear not! For it shall automatically revert them to organic organs. I tested it.
It will also prevent you from selecting them again.

### DIFF
--- a/code/modules/organs/body_modifications.dm
+++ b/code/modules/organs/body_modifications.dm
@@ -223,7 +223,7 @@ var/global/list/modifications_types = list(
 		return new replace_limb(holder)
 	else
 		return new organ(holder)
-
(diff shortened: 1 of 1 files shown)


Kazkin:
I was planning to do this already, so thank you, it saves me some time.

PR #1337, opened 2021-05-26
Defib Stat Deathloss Adjustment/Rebalancing

intense-skies:
Why I made this PR:
Currently the system punishes the weakest individuals on the server. Those who do not power game, are not robust, nor try to specifically game the system. Those who do power game, are robust, and try to game the system are the least affected.
unknown-30
The stats of an unrobust player after some time of making mistakes. They played it out through the round. Although in my eyes... At this point you may as well use the elevator and come back up with how shitty it is.
Personally an entirely better system would be made but this is a simpler and more short-term solution to an issue I see present.
What this PR does:
-Removes the Permanent -15 to all stats on defib. 
-Adds in a variable system in its place. Praise RNG.
How the variable system works is as follows:
For all stats from 0-40 you have a random loss of 1-8.
From 40 to 60 you have a random loss of 8-12
From 60 to infinity you have a random loss of 12-20 (So potentially more punishing to those of higher stats then before)
With this type of change it is a random loss with the highest number based upon the 20% of the max amount. So 20% of 40 is 8, 60 is 12, and of 100 is 20.
Nothing is done to a stat if it is below zero as you probably have more pressing issues at that point. 
Unrelated to the balance it provides a debug log so it can be seen how much the permanent amount was given and to which stat of the user.
Image of the provided logs: Capture3The test dummy has zero in all except for 100 in vig.


Trilbyspaceclone (approved):

Kazkin:
Go ahead and remove the log_debugs and replace them with admin_logs, that way we can keep track of stat drops and sickness assigns, otherwise this is a really well done PR and a much better system then what we currently have.

Kazkin:
I'll just add this to my stuff I'm gonna do today, so I'll merge it and make the changes real quick before pushing it.

PR #1365, opened 2021-05-31
Fix access to Medical's front desk.

r4d6:
## About The Pull Request
Fix access to Medical's front desk. For some fucking reason the door use Chemistry access while the windoors use Medbay General Access.


ThePainkiller:
Robos can already use their own surgery room which is exactly as geared up as medical's and it isn't as far away from it
Good to know medics are becoming more and more useless and easily replaceable instead of giving some exclusivity to them

r4d6:
> Robos can already use their own surgery room which is exactly as geared up as medical's and it isn't as far away from it
> Good to know medics are becoming more and more useless and easily replaceable instead of giving some exclusivity to them

"Hey, this guy in medical need surgery, brokens bones and shrapnels. There's no doctors, guess I'll have to do surgery as a roboticist. How about I use this room designed for surgery barely 5 meters from the patient?"

PR #1377, opened 2021-06-02
Move some FBP augment around

r4d6:
## About The Pull Request
- Replace Soteria's Combat Model's Energy Blade with a regular Armblade.
- Give SCM's Energy Blade to Blackshield's CQC Model in exchange for its shield.
- Give BQCQM's Shield to Blackshield's Ballistic Model.
- Adjust Origin Descriptions to reflect those changes.


Kazkin:
While not a bad idea, sot synths have already been nerfed quite hard with an armor and stat drop. They lost 5 armor in all areas and -5 to their combat stats. I'd rather them keep the energy blade.

PR #1443, opened 2021-06-12
Ports in Pixel Shifting

intense-skies:
This PR adds in Pixel Shifting. Usable via Ctrl+Shift+Arrow Key
Resets upon movement. Can't pixelshift while buckled.
Note: Lot of copy pasting occured here. But tested and works properly. (Not sure if it will merge easily or not though)

### DIFF
--- a/code/datums/movement/mob.dm
+++ b/code/datums/movement/mob.dm
@@ -308,6 +308,10 @@
 	for (var/obj/item/weapon/grab/G in mob.grabbed_by)
 		G.adjust_position()
 	*/
+	//Sojourn pixel shift port
+	mob.pixel_x = mob.default_pixel_x//Reset pixel shifting x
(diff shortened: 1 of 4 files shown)


MSieghardt:
It works, but it can be confusing as you can pixel shift into walls or pixel shift so that people don't know what tile you're on. You can macro it to change your position back by 12 pixels, making it so someone would miss you in melee if they tried to click you. (Example, a macro can have multiple Lines. You can do "westshift /nwestshift /nwestshift /nwestshift /nwestshift /nwestshift /nwestshift" to rappidly move to that direction pixel-wise and maybe make someone misclick.

This also allows you to hide fairly well. Also allows you to cause this kind of confussion:
image


Which tile am I on? The one with the vent or the one without? Which will you shoot?

Kazkin:
Lets see what happens fam.

PR #1491, opened 2021-06-18
Adds in a Fractal Rosary to loadout

intense-skies:
What this PR does:
-Adds in a fractal rosary to the game as well as pickable from loadout

What is a fractal rosary?
"This is an insignia given out by the Church of Absolute to people who consider themself to be a Fractal: An individual who believes and follows the Church but has not yet or cannot be inducted to full membership."

So far only it can only be acquired from loadout (might add it as something that can be acquired like cruciforms for vectors/prime to give to people.) 
Basically for those who want to play a character who wants to show their church affiliation. 

It appears as follows:
Capture
Capture2
Capture3
Sprite done by Polyushko, Gabi's Zaręczony#0323 from the discord.


Kazkin:
Nice, I like it.

PR #1530, opened 2021-06-28
Revert superior mob conversion

r4d6:
Revert all my hard work, aka https://github.com/sojourn-13/sojourn-station/pull/1518 & https://github.com/sojourn-13/sojourn-station/pull/1521

Still keep them commented out in case we want them back in.

I would honestly recommend trying https://github.com/sojourn-13/sojourn-station/pull/1531 before merging this revert, because the only reason this is being reverted is lots of lag, and Trilby's PR will probably fix that.


r4d6:
Trilby's optimization PR make this PR useless.

PR #1607, opened 2021-07-13
Adds in the Matriarch Cape (hat)

intense-skies:
Adds in a matriarch specific item (which works like a hat) for Jea Smail.
Admin only spawn in.
Sprite done by BladeburstNINJA#9500


intense-skies:
I'm retarded, pls ignore

PR #1608, opened 2021-07-13
Adds in the clothing for the Matriarch + update skinning knife icon

intense-skies:
Adds in a a Matriarch cape, admin spawn only. Fits in the hat slot since Jea normally doesn't wear one
Also provides new icon and updates the skinning knife to use it (old one still in files).
**Sprites done by BladeburstNINJA#9500**


intense-skies:
Finally did it right this time thank you god.

ThePainkiller:
Woah those are some nice stats for a "headgear". Have you thought about making it an actual attachable accessory so you can wear it over a normal suit instead?

r4d6:
> Woah those are some nice stats for a "headgear". Have you thought about making it an actual attachable accessory so you can wear it over a normal suit instead?

Well, can accessories even **give**armor ?
And in any case, this is an admin-only item designed for the leader of a faction.

Kazkin:
Given it only armors the head, I'm  fine with it.

PR #1626, opened 2021-07-17
Nerf Bluecross Hydrogen-Plasma gun.

r4d6:
## About The Pull Request
The 'Reclaimator' Bluecross Hydrogen-Plasma Gun now take 10 hydrogen-plasma per shot and lose its infinite cell.
It still doesn't heat up through.

### DIFF
--- a/code/modules/projectiles/guns/plasma/variant.dm
+++ b/code/modules/projectiles/guns/plasma/variant.dm
@@ -40,24 +40,26 @@
 /obj/item/weapon/gun/hydrogen/incinerator
 	name = "\improper \"Reclaimator\" hydrogen-plasma gun"
 	desc = "An anomalous weapon created by an unknown person (or group?), their work marked by a blue cross, these weapons are known to vanish and reappear when left alone. \
-	This plasma gun doesn't seems to heat up and its hydrogen cell doesn't seem to run out."
+	This plasma gun doesn't seems to heat up."
 	icon_state = "incinerator"
-	use_plasma_cost = 0
+	use_plasma_cost = 10
 	heat_per_shot = 0 // No heat gain.
 	origin_tech = list(TECH_COMBAT = 15, TECH_MATERIAL = 7, TECH_PLASMA = 25)
 	matter = list(MATERIAL_PLASTEEL = 65, MATERIAL_MHYDROGEN = 15, MATERIAL_OSMIUM = 10, MATERIAL_TRITIUM = 5)
 	init_firemodes = list(
-		list(mode_name = "standard", projectile_type = /obj/item/projectile/hydrogen, fire_sound = 'sound/weapons/lasercannonfire.ogg', fire_delay = 30, icon = "destroy", heat_per_shot = 0, use_plasma_cost = 1),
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/plasma/variant.dm
```diff
@@ -40,24 +40,26 @@
...
-		list(mode_name = "standard", projectile_type = /obj/item/projectile/hydrogen, fire_sound = 'sound/weapons/lasercannonfire.ogg', fire_delay = 30, icon = "destroy", heat_per_shot = 0, use_plasma_cost = 1),
-		list(mode_name = "overclock", projectile_type = /obj/item/projectile/hydrogen/max, fire_sound = 'sound/effects/supermatter.ogg', fire_delay = 50, icon = "kill", heat_per_shot = 0, use_plasma_cost = 1)
+		list(mode_name = "standard", projectile_type = /obj/item/projectile/hydrogen, fire_sound = 'sound/weapons/lasercannonfire.ogg', fire_delay = 30, icon = "destroy", heat_per_shot = 0, use_plasma_cost = 10),
+		list(mode_name = "overclock", projectile_type = /obj/item/projectile/hydrogen/max, fire_sound = 'sound/effects/supermatter.ogg', fire_delay = 50, icon = "kill", heat_per_shot = 0, use_plasma_cost = 20)
 	)
 
 /obj/item/weapon/gun/hydrogen/incinerator/Initialize()
 	..()
-	flask = new /obj/item/weapon/hydrogen_fuel_cell/infinite(src)
+	//flask = new /obj/item/weapon/hydrogen_fuel_cell/infinite(src) // Apparently never running out of ammo is too OP.
```
ThePainkiller, line 56: Instead of outright commenting out the cell, why not edit `hydrogen_fuel_cell.dm` to make the "infinite" cell  just be a 500 plasma capacity one that DOES get its plasma consumed with every shot? That way the gun has a large shelf life to make it worthwile instead of being limited even more by the nicheness of its ammo system?
(Another suggestion as well, add them hydrigen fuel cells to the random loot pool, like normal cells...)

r4d6: Well, that way if the admins need an infinite hydrogen cell for any reasons, they can spawn it in.
Beside, if I **do** make the cell non-infinite with 500 shots, then that would just be a smaller and better Backpack Cell.
And if I don't allow the cell to be taken out, Then that mean it can't be reloaded.

Also, why would I remove the gun's ability to endlessly shoot, but give it near-endless shots ?

ThePainkiller: Because it's a bluecross weapon, who is tied on rng spawn among an ever increasing pool of bluecross spawns and supposed to be a very rare and strong weapon variant of normal weaponry or its own thing, that only lasts for one shift in someone's hand and is never guaranteed?
Couple that with the fact science is the only way to obtain its ammunition and the niche narrows even further, might as well fade it into obscurity altogether
500 ammo with a 10/20 cost per shot means 50 shots, which is acceptable for a Bluecross weapon whose ammo you can't scrounge or refill anywhere but science and only if it's node is available

r4d6: Soteria can't make the ammo, you buy it at Lonestar for 1$ per unit of hydrogen, so a crate of 20 flasks for 300 shots or a crate with a single backpack-cell that hold 500 shots.

PR #1631, opened 2021-07-18
Add an hydrogen generator

r4d6:
## About The Pull Request
Basically a pacman that use Hydrogen Fuel Cell to make power.

It can hold a single hydrogen fuel cell at a time.
It is also upgradable, with the capacitors managing how much power it produce, from 50 kW at T1 to a whooping 300 kW at T5.


r4d6:
So I got a choice to make : To make the generator only able to hold a single flask, or to keep its six slots.
On one hand, being able to hold multiple flasks is good.
On the other, having only 1 slot would make thing easier to code & neater.

PR #1652, opened 2021-07-24
Add Tesla Turret

r4d6:
## About The Pull Request
Add a new build, the Tesla Turret !

Needing to be built on a power cable (same as pacmans) and with a **vertical** frame, this bad boy will proceed to shock every set target in a 5 tile radius!

It use a capacitors as components, the better the capacitor, the faster it can zap !

You can change the target using a multitool, or anything else with the PULSING quality.

Current Choices are : 
- Humanoids (AKA players)
- Silicons (AKA cyborgs)
- Superior Mobs (AKA Roaches, spiders, robots, e.t.c...)
- Simple Mobs

One thing to consider is that it **doesn't care** who it is zapping, so it can, and will, zap those handmade bots as much as it does the roaches and spiders.

Currently with placeholder sprites, waiting on Nayu to finish better sprites that fit with our graphical style.


Kazkin:
Want me to wait on this until we get the sprites?

r4d6:
> Want me to wait on this until we get the sprites?

Yes please.

PR #1683, opened 2021-08-02
Add improvised multitool implant

r4d6:
## About The Pull Request
Port https://github.com/discordia-space/CEV-Eris/pull/6370
Add a craftable multitool implant that use junk tools.

Also make the handmade roomba clean the tiles it pass over again.


code/datums/craft/recipes/tools.dm
```diff
@@ -77,6 +77,26 @@
...
 	)
 
+// For when you want everything in a single arm.
+/datum/craft_recipe/tool/toolimplant
+	name = "Multitool implant"
+	result = /obj/item/organ_module/active/multitool/improvised
+	steps = list(
+		list(/obj/item/storage/toolbox, 1),
+		list(/obj/item/circuitboard, 1),
+		list(/obj/item/organ/external/robotic, 1),
```
ThePainkiller, line 87:
```suggestion
		list(/obj/item/organ/external/robotic/junktech/, 1),
```
IMO it makes more sense for a makeshift tool implant with improvised tools to take an improvised junk organ

ThePainkiller (changes requested):

code/datums/craft/recipes/tools.dm
r4d6: The way it is currently set up, _**any**_ robotic limb, not only junktech, but any, can be used.
Restricting it to only junktech would just add even more step, not to mention that people would probably just make junktech limbs anyway.

ThePainkiller: Fair, I just wish it was the same for the junktech robot to make since it requires a **borg** arm and thus uncraftable through scavenging as most machines drop implants and not borg parts...

PR #1699, opened 2021-08-05
Change the roboticist loadout

r4d6:
## About The Pull Request
Since Soteria is all about high-tech stuff, give Roboticists a bit of a new loadout consisting of :
- Electric Screwdriver
- Big Wrench
- Hydrogen-Plasma Welder
- Normal Crowbar
- Normal Wirecutters
- Normal Multitool
- RoboAnalyzer

In addition, it move the cable coil from their belt to their bag, with their analyzer going the opposite way, and adding an hydrogen fuel flask to their bag too. Wouldn't want them to be unable to use their fancy new welders.

*Also I used this occasion to fix a few things with the hydrogen welder, in preparation of allowing them to have toolmods installed.


ThePainkiller:
tbh I'd put the cable coil on the toolbelt and the robot analyzer in the backpack, much like how doctors start with health analyzers on them, nobody keeps them on their EMT belt!

PR #1709, opened 2021-08-07
Cargo Overhaul

Kazkin:
-Cargo crates now have a surplus tag when ordered, this causes anything ordered to be worth 10% of its actual value. This change means we no longer need to do extensive math to balance price and value and can just add things to cargo as rule of cool.
-Moved var/price to cargo supply for logics sakes.
-Surplus tag is now viewable by those with the market prof perk to prevent scamming and make it cleaner/move obvious.
-Guns now sell for displayed value instead of less.
-Wandering trading vessels which may or may not spawn now offer various trade goods to buy or sell, refreshing every 25 minutes. The prices can be a bit wonky but can be worth purchasing. This system will be tweaked as needed.
--These ships have a massive variety of goods that will appear and disappear in addition to what they may buy. Those seeking active buying and selling may do so by linking an account and either ordering the items on the short range teleporter or sending off the correct amount of items beacon to earn the listed price to their account.
-Overhauled alot of cargo prices given the new surplus system, allowing for much cheaper items for sale.
-Removed most rifles from cargo, added more pistol types, and added some novelty kit crates for fun/money spending.


Kazkin:
Why the fuck are their conflicts >:|

r4d6:
> Why the fuck are their conflicts >:|

It seems to conflict with my PR that messed with the mech's guns.

Sorrey !

PR #1742, opened 2021-08-12
Allow chem dispensers to start hacked.

r4d6:
## About The Pull Request
This allow for mapped-in chem dispenser to start hacked, like the Industrial Chem Dispenser in Soteria.


ThePainkiller:
Here's a tip: **ALL** mapped chemical dispensers (yes this includes booze and soft drinks) start hacked on the map due to a var oversight, but since they're `new` they get initialized with their standard base chems `list()`, requiring you to use a multitool TWICE on them to unlock their "contraband chems". Does this fix that issue? Do they spawn with their corresponding "hacked" chemicals now?

r4d6:
> Here's a tip: **ALL** mapped chemical dispensers (yes this includes booze and soft drinks) start hacked on the map due to a var oversight, but since they're `new` they get initialized with their standard base chems `list()`, requiring you to use a multitool TWICE on them to unlock their "contraband chems". Does this fix that issue? Do they spawn with their corresponding "hacked" chemicals now?
> What I mean is, is it tested?

I tested it with Soteria's Industrial Chem Dispenser, which is what the map change of this PR is about, and it *does* work, however it doesn't change the rest of the dispensers as I don't know which one is intended to start hacked.
And instead of hard-coding that all chem dispenser start hacked if they're mapped, I do think it would be better for it to be on a case-by-case basis, as I don't doubt that someone people will map chem dispensers that aren't intended to start hacked.

ThePainkiller:
> I tested it with Soteria's Industrial Chem Dispenser, which is what the map change of this PR is about, and it _does_ work, however it doesn't change the rest of the dispensers as I don't know which one is intended to start hacked.
> And instead of hard-coding that all chem dispenser start hacked if they're mapped, I do think it would be better for it to be on a case-by-case basis, as I don't doubt that someone people will map chem dispensers that aren't intended to start hacked.

Like I said, **ALL** of them are. Every single one.
Soda dispensers (and the booze dispenser) at Greysons? Pre-hacked
Same for those machines colony side
Chemical dispenser on Void Wolf base and Forgotten bunker? Pre-hacked
You craft a new dispenser by hand? Pre-hacked, have to use multitool twice
The `fancy_hack` var is defined as false on the parent chem dispenser, and some of the booze/soda machines have it set to `true` for some reason, yet all of them start mapped as hacked and you have to multitool them twice for some reason...

PR #1743, opened 2021-08-12
Update the folken healing perk

r4d6:
## About The Pull Request
Every folken now start with the Folken Healing Perk, which will fix Ethnicity-less folken from burning in the colony.
The Young Folken Healing Perk now replace the normal folken healing perk and put it back if it is lost for any reason.


ThePainkiller:
Honestly since we're at it, their healing in the dark perk needs a nerf, it's almost like back in the days of tg's old virology in terms of speed, should be at par with others like Naramad healing through water or human racials to not powercreep them entirely

r4d6:
> Honestly since we're at it, their healing in the dark perk needs a nerf, it's almost like back in the days of tg's old virology in terms of speed, should be at par with others like Naramad healing through water or human racials to not powercreep them entirely

Fixing the mycus' dark healing is outside of this PR scope of simply making every folken have the light healing perk, if only to make them stop burning if they forgot to set an ethnicity.

PR #1772, opened 2021-08-20
Code cleanup + fixes

ThePainkiller:
## About The Pull Request

Eliminates some unnecessary and repetitive bits of code for improvement

**NOTE:** Certain macros are conflicting with pre-defined ones (Artist Bench and Tesla Turret), but I'd rather ask for confirmation before touching those two as I have no idea if the artist bench will properly call for the adequate caliber macros when crafting. Tesla turret is just merely a matter of renaming macros to something that does not conflict, but as those get undefined by the end of file they shouldn't pose a real conflict.
Please comment below if you're better informed than I am on these two macro defines so I don't accidentally go deleting stuff that actually works. Specially the Artist's bench.

- Kvass can now be crafted by combining Flour and Ethanol in equal parts in presence of 5u of Universal Enzyme
- Mead is now crafted with water and honey instead of sugar (1 part Honey, 1 part Water, in presence of 5u Universal Enzyme), blacklist its reaction from the microwave to prevent conflicts with Ethanol reaction and Honey Buns recipe
- Fixes Crevice Spike having an invisible glass sprite due to icon state missmatch
- Waffles no longer require a butter slice to not conflict with Croissants
- Icecream sandwiches now require brownie slices to craft instead of cookies. Strawberry icecream sandwich recipe unchanged.

## Changelog
:cl:
add: Adds recipe for homemade Kvass: 1 part ethanol + 1 part flour in presence of 5u Universal Enzyme
tweak: Mead is now made with equal parts Water and HONEY in prescence of 5u Universal Enzyme, to not conflict with Ethanol recipe. Also blacklists this reaction from the microwave to not conflict with Honey Buns recipe
tweak: Icecream sandwich now crafted by attacking the icecream with a Brownie Slice. Strawberry Icecream Sandwich recipe unchanged.
fix: Crevice Spike glass should no longer be invisible
fix: Waffles no longer require Butter Slice to cook, to not conflict with croissants
code: removes unnecessary and repetitive tidbits, macros previously defined, indentatation issues
/:cl:

### DIFF
--- a/code/__DEFINES/qdel.dm
+++ b/code/__DEFINES/qdel.dm
@@ -24,10 +24,6 @@
 #define QDEL_NULL_LIST(x) if(x) { for(var/y in x) { qdel(y) } ; x = null }
 #define QDEL_CLEAR_LIST(x) if(x) { for(var/y in x) { qdel(y) } ; x = list() }
 
-#define GC_QUEUED_FOR_QUEUING -1
-#define GC_QUEUED_FOR_HARD_DEL -2
-#define GC_CURRENTLY_BEING_QDELETED -3
-
 /**
  * Delete `item` after `time` passed.
  * Return `id` for timer, so deletion process could be stopped.
--- a/code/game/machinery/autolathe/artist_bench.dm
+++ b/code/game/machinery/autolathe/artist_bench.dm
@@ -386,4 +386,4 @@
 #undef ERR_NOLICENSE
 #undef ERR_PAUSED
 #undef ERR_NOINSIGHT
-#undef MAX_STAT_VALUE
+
--- a/code/game/objects/items/weapons/clown_items.dm
+++ b/code/game/objects/items/weapons/clown_items.dm
@@ -9,7 +9,7 @@
  * Banana Peals
  */
 //its here so that its all contained in 1 file for the var
- /obj/item/bananapeel
+/obj/item/bananapeel
  	var/how_many_times_we_can_pull_a_pro_clown_gamer_move = 1
 
 /obj/item/bananapeel/Crossed(AM as mob|obj)
--- a/code/game/objects/items/weapons/dna_injector.dm
+++ b/code/game/objects/items/weapons/dna_injector.dm
@@ -418,7 +418,6 @@
 /obj/item/dnainjector/midgit/New()
 	..()
 	block = SMALLSIZEBLOCK
-		..()
 
 /obj/item/dnainjector/antimidgit
 	name = "\improper DNA injector (Anti-Small Size)"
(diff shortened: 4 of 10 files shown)


r4d6:
I wasn't aware that the Tesla Turret defines were conflicting...

ThePainkiller:
> I wasn't aware that the Tesla Turret defines were conflicting...

They're not _"conflicting"_ per se, just getting a warning about them being macros previously defined in `mob.dm` as bump flags, but like I said, since they're undefined by the end of it, it shouldn't conflict...

r4d6:
> > I wasn't aware that the Tesla Turret defines were conflicting...
> 
> They're not _"conflicting"_ per se, just getting a warning about them being macros previously defined in `mob.dm` as bump flags, but like I said, since they're undefined by the end of it, it shouldn't conflict...

[Better safe than sorry.](https://github.com/sojourn-13/sojourn-station/pull/1773)

ThePainkiller:
> [Better safe than sorry.](https://github.com/sojourn-13/sojourn-station/pull/1773)

Welp, saves me one commit! Thanks!

Kazkin:
Is this ready to be submitted?

ThePainkiller:
> Is this ready to be submitted?

Compiles and is semi ready, I wanted your opinion on removing the firearm caliber macros defined on the Artist's bench since they are already defined on `_defines.dm` and **should** draw from them. I have no way of testing since nanoui does not work on localhosting for me if it'd work, only if it can compile.
If yes, I'll just make another commit eliminating those defines from the artist's bench and then it'd be ready

Kazkin:
> 
> 
> > Is this ready to be submitted?
> 
> Compiles and is semi ready, I wanted your opinion on removing the firearm caliber macros defined on the Artist's bench since they are already defined on `_defines.dm` and **should** draw from them. I have no way of testing since nanoui does not work on localhosting for me if it'd work, only if it can compile.
> If yes, I'll just make another commit eliminating those defines from the artist's bench and then it'd be ready

This was actually a critical error from what I remember when using the artist bench originally, as our code was different from Eris since they use simple calibers (like 4 types) and for some reason would not play nice with our defines unless defined and then undefined specifically in that file. It also limited the caliber types, as originally you could get revolvers chambered in rpg rockets.

Assuming you removed the defines and it compiled, plus if memory serves caliber type checks were added in the creation step, you *should* in theory be fine, though I'd say test it to be sure by randomly spawning artist revolvers since your NanoUI is buggy.

ThePainkiller:
Nothing I've tried either compiles nor works for the Artist's Bench, which is incredibly odd considering it uses `R.caliber` var to call for `CAL_PISTOL` and etc, so I don't understand why the bench would **NEED** those defined in itself to work...
I'll try revisiting it up later, in another PR, this one is good to merge as-is, compiles and all

PR #1782, opened 2021-08-20
Add Greyson Lathe

r4d6:
## About The Pull Request
Add the greyson lathe, who's sprites were made by HOUND, which has 4 manipulators and four lasers, which is 2 more than the industrial lathe, but still have 6 matter bins.

This, combined with a lowered minimum cost ceiling, mean that with T4 parts or better, you can print items up to 5% of its cost instead of 20%.

So a shovel costing 5 steel to print cost 1 steel to print with T3 parts and 0.25 steel with T5 parts.


Kazkin:
Will work shop this heavily later.

PR #1801, opened 2021-08-21
Make the durand's AC2 start loaded

r4d6:
## About The Pull Request
Why would the Durand's Scattershot start loaded but not its AC-2 ?


ThePainkiller:
I can confirm I've snagged the Durand from Forgotten bunker and its AC ended up loaded with a random ammount of ammo all three times I've reached it, same as its scattershot, are you sure you're not just the victim of bad RNG?

PR #1807, opened 2021-08-22
fbp changes!

cdb-is-not-good:
"What does this do?"

A few things! First and foremost, the white "human" base for FBPs is changed to a more skeletonized 'robot' base. The intent is for this new sprite to be 'low profile', such that it can have markings overlaid on it without obtrusiveness.

"What is the point of that?"

As of now we can't select groin/torso/head prosthesis in the augments tab. This means that while your plastic person may be able to have prosthetic limbs, those three limbs will always be attached to the 'base'. With this change, and the added new markings - the intent is to essentially allow FBPs to have more freedom in the base from which they are working.
This includes a set of four markings(Ward Takehashi and Hesphaistos alt) and some new prosthetics, using sprites that we had as well as a few from Virgo.

"What if i want a human appearing fbp specifically? "

With just a single extra marking this is easily possible. This also means that - say you want non 'human' shapes to work as your base? You could easily have this, particularly with how markings respect order and 'layer'. This pull request is frankly a first draft and i expect it to be changed if it even is going to be accepted.

"Does this actually change how FPBs work in any way?"

No. The only thing it changes is their 'base'. The white human form is swapped, and new markings and prosthetics are added as proof of concept.  If you want specific race forms or such added, lemme know!
As of now, the markings are specifically bundled just into "torso/groin"+"head", but a full body form would be fine.

"Won't this further clutter the markings popup?"

Yes. I believe there is a way to denote markings as being only for certain races but i am uncertain if/how it works - it doesn't seem to.
I may also sit down and try to put together a unified way of organizing markings in the popup that is a little more intuitive than what we have now where it's just sorta - all over the place.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


cdb-is-not-good:
Whoop this aint right. Gonna figure out what i broke then try again

PR #1895, opened 2021-09-08
Add Nanite Rig Power

r4d6:
## About The Pull Request
Add more powers to the nanogate, this time it is ARMOR.

- Install Nanite Rig (3 points) : Cannot be used if you're wearing anything on your back. This power create a new rig that _**cannot be removed**_ on your back. It start with an 14000L and can be upgraded by spending more points. The rig slowly recharge over time, though constant use may empty the battery.
- Upgrade Rig Storage (1) : Require a Nanite Rig. This one add a storage compartment to your rig, allowing it to act as a backpack.
- Install Laser Cannon (1) : Require the Rig. This power install the mounted laser cannon module on your rig, allowing you to shoot by pressing the middle-mouse button, or any other button you set it to.
- Install Autodoc (1) : Require the nanite rig. This upgrade install an autodoc module on your rig, which can perform surgery. I am not 100% sure if they fully work, but they should. Don't worry, it is not the capitalist variant that drain you account fund for every action.

Also, the reason why it can't be removed is because it is literally attached to your spine.


Dongels:
> Should make non opifex get charged credits :v   yayakiki

PR #1932, opened 2021-09-17
Outsider ANOMALY HUNTERS!

intense-skies:
Special training or experience from living in the wild, untamed world has provided the instinctive advantage to know the effects of greater oddities.
-Gives outsiders the anomaly identification perk (like what prospectors have) 
-Adjusts the description of the perk to better fit this change

### DIFF
--- a/code/datums/perks/job.dm
+++ b/code/datums/perks/job.dm
@@ -409,7 +409,7 @@
 
 /datum/perk/stalker
 	name = "Anomaly Hunter"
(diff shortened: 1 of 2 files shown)


Kazkin:
Sadly, while reasonable, outsider is meant to be hard mode and giving them any sort of buff is HERESY!

In all seriousness, while an okay idea in theory, it removes the one quality that prospectors have unique to them and I'd rather not make that change.

PR #1936, opened 2021-09-18
Cry-o me a river updoot!

LongSleeves:
Nerfs base chems (bicard + dylo) by 50%
Fixes the == symbols used for OD check on ossisine and kyphotorin
Nerfs cryo and cron to be less effective. (They healed 50+ damage a second when put together. Fully healing people in 3 to 4 seconds when tubed.)


intense-skies:
The_kind_of_tired_that_sleep_won't_fix

PR #1972, opened 2021-09-25
Add filter effects, holy glow to litanies and more.

r4d6:
## About The Pull Request
- Add a proc to easily add timed [Filter Effects](http://www.byond.com/docs/ref/#/{notes}/filters) to mobs.
- Most of the Church's litanies now make the affected people glow for 2.5 seconds.
- Removed the 'anonymous' part of the message litany because it wasn't.
- Change the Factorial's mass repair litany to no longer affect the Factorial. You got a separate litany for that, use it.

### DIFF
--- a/code/__HELPERS/filters_effects.dm
+++ b/code/__HELPERS/filters_effects.dm
@@ -1,0 +1,27 @@
+/* 	This is for adding & removing filter effects from mobs. (See : http://www.byond.com/docs/ref/#/{notes}/filters )
+	[user] is the mob to add the filter effect to.
+	[effect] is the filter effect you want to add.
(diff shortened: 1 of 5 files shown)


intense-skies:
Please only add the glow to the following litanies:
- AoE Healing
- AoE Buffs
- On the subject of direct healing (the one where you lay your hands onto a fellow disciple)
- On the subject who gets their cruciform activated/reactivated
As someone who plays the church a lot please take this suggestion into consideration. They already have a lot of audio and visual thus far.

Kazkin:
Got to agree, the glow should only effect AoE litanies, as its good feedback and more thematic than making every litany make you glow.

PR #1977, opened 2021-09-26
More Craftable Leather Storage

intense-skies:
What this PR does:
-Adds in leather pouches for small, med, and large.
-Allows them to be crafted much like the rest of the leather storage items such as backpacks, satchels, and duffles
Primarily aimed at allowing hunters to have some more storage options instead of using TRASH BAGS as if they were a garbage sifting outsider
Capture
Picture of the pouches. Crafting goes 1 leather for small, 2 leather for med, 4 leather for large. Plus the usual extra crafting steps.
**Sprites done by BladeburstNINJA#9500**


Dongels:
Fuck those sprites look kick ass! *thumbs up*

cdb-is-not-good:
super good sprites but like. why? this means that lonestar, soteria, and the lodge can now consistently produce pouches. Yea it's a dumb monopoly to have - but why bother taking it away from them at all?

r4d6:
> super good sprites but like. why? this means that lonestar, soteria, and the lodge can now consistently produce pouches. Yea it's a dumb monopoly to have - but why bother taking it away from them at all?

I mean they already could make backpack out of leather.
The only monopoly is people annoying the church to get pouches. And even then that monopoly will still stay as the Church can (I think) make the other types of pouches like ammo or tubular pouches, not to mention that having a gardener or someone working Xenobotany isn't guaranteed, and so not reliable.

PR #1981, opened 2021-09-27
Removes the Litany Glow

intense-skies:
Reasoning behind this is that it both looks bad and everyone who I've spoken to or experienced them has had a negative opinion regarding it
scrnshot1
Removing for the sake of all church and colony members who do not deserve an aura of piss.


intense-skies:
Closing for now.

PR #2111, opened 2021-10-13
Make Soteria Great Again

hex37:
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
Rebalances the SST's that require Large cells to take a less insane amount of charge per shot. Change comes on the grounds that the massive batteries take up way more storage space than regular mags of ammo and should be recognized for their unwieldy chonk.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/sst.dm
+++ b/code/modules/projectiles/guns/energy/sst.dm
@@ -54,7 +54,7 @@
 	wielded_item_state = "_doble"
 	icon_state = "humility"
 	item_state = "humility"
-	charge_cost = 800
+	charge_cost = 500
 	fire_sound = 'sound/weapons/guns/fire/shotgunp_fire.ogg'
 	gun_tags = list(GUN_PROJECTILE, GUN_LASER, GUN_ENERGY)
 	w_class = ITEM_SIZE_HUGE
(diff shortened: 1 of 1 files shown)


MSieghardt:
To answer your points:
Two bullets is enough to put any mob to sleep from the System Cost. Three to do it faster. 
The weapon is designed to be used against players, not mobs, where even one bullet can guarantee a down due to slowdown.
As for the cells. It takes 40 seconds to recharge an 8000L cell when using Greyson capacitors (which Soteria can get if they know what they're doing). Also, why use such a small cell, you're in the department that makes cells.

It has decent accuracy and the slow fire rate is intentional. You don't want to kill your target, just incapacitate. 

I recommend against merging this. 

To me this sounds like:
I wanted to be prospector++ but this gun is only good against players.

hex37:
The only thing this does is make a gun more economical for use. Realistically speaking, if something isn't dead by the time you fire 16 of these things, it probably won't be going down in the second clip. This just makes it take less ammo when in use.

Even if R4D6's change gets implemented, I'd still push for this addition to increase its viability as a late-stage weapon SI can supply to Trauma Team.

MSieghardt:
In the end, I think you're right that we should reduce the energy cost of it. I guess this is okay.

PR #2112, opened 2021-10-13
Replace the Trauma Team's LMG with a Bullpip

r4d6:
## About The Pull Request
Replace the Trauma Team's SST LMG with a Bullpip.
The Trauma Team's main object require them to fight non-colonist hostiles, and while the LMG is great against a single target or three, it's high energy consumption and low damage mean it is unsuited for facing hordes of roaches or spiders.
As such their LMG is now replaced by Soteria's Bullpip .35 SMG and 3 magazines, with RnD being able to print more with a bit of research. The Trauma Team still keep their Format-Bound to quickly incapacitate colonists and other people.

### DIFF
--- a/code/datums/outfits/jobs/medical.dm
+++ b/code/datums/outfits/jobs/medical.dm
@@ -46,8 +46,8 @@
 	back = /obj/item/storage/backpack/duffelbag
 	l_pocket = /obj/item/device/flash
 	r_pocket = /obj/item/handcuffs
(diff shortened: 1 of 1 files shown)


cdb-is-not-good:
the Bullpip is great across the board - but downgrading trauma team isnt necessarily a bad thing. they have arguably one of the b est RIGs in the game right now and solid round-start weapons.

PR #2205, opened 2021-10-31
Re-add the antimatter engine

r4d6:
Reverts sojourn-13/sojourn-station#1714

Even if I wasn't done with the changes, it doesn't affect anything in the game currently, and since I didn't keep the branch I made the first PR on, I need the revert if we want to work on that again.

For those who don't know, the Antimatter Engine is basically a big, scalable PACMAN that use fuel that you buy from cargo. The bigger it is, the more power it can produce at the cost of using more antimatter.

Currently, each core can safely support 2 unit of antimatter each, every unit of antimatter injected produce 300 kW.
Injecting more antimatter than the cores can support is possible and will result in more power, but the stability will drop, and at 0 stability, it explode.

The cargo crates are commented out until an event arrive to introduce it.


Trilbyspaceclone (changes requested):
image
Your removing soul form the game. 2/10

r4d6:
> image You're removing soul form the game. 2/10

Well at least it is similar-looking. And it is Citadel that changed it, not me. I just ported the sprites because all we had was a literal circle.

PR #2243, opened 2021-11-05
Fixes nanite regen to work on synths (as wiki intended)

Yawet330:
Title

## Changelog
:cl:
fix: Nanogate regen now applies to synthetics, as implied by the wiki.
/:cl:


MSieghardt:
This would make FBPs, which are already very strong and have 0 downsides to a nanogate, a bit too strong in its state. If fixed to work as heal_overall_damage does, that would probably be fine, but giving them an enhanced healing version is just too strong.

Yawet330:
> This would make FBPs, which are already very strong and have 0 downsides to a nanogate, a bit too strong in its state. If fixed to work as heal_overall_damage does, that would probably be fine, but giving them an enhanced healing version is just too strong.

This was discussed in dev-discussion.

Kazkin:
Slow the regen to 1 healing per 5 ticks, slower given how powerful synths/synthetic parts are.

PR #2489, opened 2021-12-23
Makes obligate name = false the default.

DrPainis:
## About The Pull Request
<details>
<summary>
Only races with obligate name = true in their code actually have obligate names now.
I literally only did this because I want the spider upbringing on a character that is supposed to be a custom species and not a cht'mant.
</summary>
<hr>
	
<hr>
</details>

## Changelog
:cl:
tweak: most species can now be renamed
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species.dm
+++ b/code/modules/mob/living/carbon/human/species/species.dm
@@ -15,7 +15,7 @@
 
 	var/default_form = FORM_HUMAN	//If nothing else sets it, what do we look like.
 	var/obligate_form = FALSE		//If true, character creation will force the use of either this form or its subforms.
-	var/obligate_name = TRUE		//If true, forces the character's species name and name color to conform.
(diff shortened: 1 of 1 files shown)


MSieghardt:
I'm gonna go ahead and say 'no' to this one. This will ruin a lot of stuff in terms of how people interpret races. If you want to be a cht'mant, you need to be a cht'mant, you can't be a snowflake race and have the cht'mant / other races benefits and changes.

A good example would be someone giving you iron to increase your blood regeneration yet ending up killing you because they didn't know you're a cht'mant. 

This applies to other races too.

Kazkin:
For balance reasons, this has to be denied, as it would make it impossible to tell another persons race for malign or noble intentions. The immediate problem is, being unable to tell someones race due to its custom name, is how certain species react to healing chemicals and damage types would be difficult, if not impossible. There are other reasons, but this is sadly a no.

DrPainis:
Just make it so the color indicates race instead.

PR #2594, opened 2022-01-10
Carp stun improvements

Yawet330:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
fix: Adjusts carp code to respect stats
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/carp.dm
+++ b/code/modules/mob/living/simple_animal/hostile/carp.dm
@@ -13,7 +13,6 @@
 	speed = 4
 	maxHealth = 25
 	health = 25
-	var/knockdowncapable = TRUE //Fix
 
 	armor = list(melee = 5, bullet = 10, energy = 50, bomb = 25, bio = 100, rad = 25) //Lasers dont work on scales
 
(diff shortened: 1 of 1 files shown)


Yawet330:
Carps have a 15% chance to paralyse you regardless of toughness/robustness, hence why this was done.

Just no.

MSieghardt:
All carps should have the chance to knock you down. Maybe just lower the chance for the baby ones? But otherwise, it's their one thing.

Yawet330:
> All carps should have the chance to knock you down. Maybe just lower the chance for the baby ones? But otherwise, it's their one thing.

Fair point; as iterated in code-dev, however. Baby carp are far more common, and far more threatening in groups due to their small size (making them harder to click), along with decent HP which makes using ballistics against them either pointless, or outright wasteful.

I might make this scale with all carp instead, based off of toughness, just having a flat-out toughness roll/check to prevent the knockdown (as it also should be for roaches/etc)

It just feels extremely unfair at times, just because of how numerous they are and the fact they use paralyse()

Yawet330:
Carp now do the following math to check if they should knock someone down

Prob(65) - STAT_TGH

This makes it so unless you have less than 65 toughness, you cannot be knocked down by carp (encouraging naturally tough; and lategame characters to spaceloot), along with  helping fix how unfair they can feel

For example, at 40 toughness (the average of a roundstart character focused towards toughness, or a naturally tough race); you only have ~25% chance to get knocked down per attack.

TECHNICALLY a buff to carp.

Yawet330:
Merging with trilbys PR

PR #2665, opened 2022-01-28
Allows you to over-ride the mech DNA lock with a very hard MEC skill

MSieghardt:
Allows you to over-ride the mech DNA lock with a very hard MEC skill that takes very long.
tested. it works.


r4d6:
Nice

PR #2678, opened 2022-01-30
Removes sliding from C, makes it so users have to macro it instead.

Yawet330:
This was a stupid change.

Placing a key that STUNS YOU AND FORCES YOU TO DROP YOUR WEAPONS right next to the KEY TO SWITCH BETWEEN YOUR HANDS is a FUCKING AWFUL IDEA

Eris designers should be shot in the head.

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -167,12 +167,6 @@ macro "borghotkeymode"
 	elem 
 		name = "CTRL+Y"
 		command = "Activate-Held-Object"
-	elem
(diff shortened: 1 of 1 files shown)


Trilbyspaceclone (changes requested):
This breaks sliding

Yawet330:
> This breaks sliding

no it doesnt, you can directly call the commands via macro

Kazkin:
People will get used to it.

PR #2758, opened 2022-02-16
Make the Tesla Turret cap it's power input instead of damage output

r4d6:
## About The Pull Request
Make the Tesla Turret cap it's power input instead of making it waste excess power by capping the damage output only.


r4d6:
Closing because apparently @hex37 has to rewrite all of the power code.

PR #2763, opened 2022-02-17
Tesla turret overhaul

hex37:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
Makes the tesla turret universally better in every way, it:
-Can be destroyed
-Can be Emagged
-Can target all kinds of animals at once
-Has turret networks for easy management
-Has an integrated UI
-Can check records and apprehend wanted suspects
-Can be set to attack anyone without a valid ID
-Can be set to target anyone without access
-Can be set to purge the robots
-Can be set to lock out AI
-Draws from an APC, meaning it doesn't need to be wired to function
-Does your taxes
-Can be hacked and subverted by the enemy
-Has a new (badly made) destroyed state
-Has better target acquisition, and doesn't fire at things it can't realistically target
-Can be subdivided into up to 5 different turret networks that can each have their own configuration
/:cl:


code/game/machinery/tesla_turret.dm
```diff
@@ -1,137 +1,619 @@
...
+
+/obj/machinery/tesla_turret/Topic(href, href_list)
+	if(..())
+		return 1
 
-	if(active)
-		to_chat(user, SPAN_NOTICE("You can't work with [src] while its running!"))
+	if(href_list["command"] && href_list["value"])
+		var/value = text2num(href_list["value"])
+		if(href_list["command"] == "enable")
```
r4d6, line 184: It would probably be better to do a `switch` statement here, instead of an if-else train.

code/game/machinery/tesla_turret.dm
```diff
@@ -1,137 +1,619 @@
...
+		return TURRET_NOT_TARGET
+
+	if(!colony_allied_turret && L.colony_friend) //If were not allied with the colony we attack them and their pets
+		return TURRET_SECONDARY_TARGET
+
+	if(colony_allied_turret && !L.colony_friend) //If were allied with the colony and we attack things that are not are pets
+		return TURRET_SECONDARY_TARGET
+	*/
+
+	if(issilicon(L))
```
r4d6, line 525: So it will shoot at synths even if they aren't too far away?

r4d6 (commented):
Looks great overall!

However I'm also seeing multiple 1s & 0s that could be replaced with TRUE and FALSE.

code/game/machinery/tesla_turret.dm
hex37: syntactically identical but it's fine either way

code/game/machinery/tesla_turret.dm
hex37: Fair enough: I can fix that.

PR #2801, opened 2022-02-24
The Revenge of the Trilogy: Risikio's Own ATK Ideas

Risikio:
**Locks the Advanced Trauma Kit in the following ways**

>Using an ATK in actual surgery to heal organs is locked to: Doctors, Roboticists, and Soteria Heads

>Using an ATKs and ABKs in the field to heal Brute/Burn damage like bandages/ointment is locked to: Sawbones, Corpsman, Trauma Team, Doctors, Roboticists, Soteria Heads

> Choosing the Salvager Alternate Title "Sawbones" now removes the _Expert Scavenger_ perk and replaces it with the _Medical Expertise_ perk, allowing for the use of ATKs and BTKs in the field.

**Pros:** 
Not having to rip out hundreds of items and recode everything.

Makes surgery and big boy injuries really only possible to be repaired by those who have proper Soteria Medical training.

Allows Corpsman, Trauma Team, and Sawbones to still be big healers in the field long enough to get them to a proper medical professional.

Sets up an area in code for future "Tweaks" to specific alternate titles.

### DIFF
--- a/code/controllers/subsystems/jobs.dm
+++ b/code/controllers/subsystems/jobs.dm
@@ -327,6 +327,8 @@ SUBSYSTEM_DEF(job)
 			for(var/datum/gear/G in spawn_in_storage)
 				G.spawn_in_storage_or_drop(H, H.client.prefs.Gear()[G.display_name])
 
+		job.alt_TitleTweaks(H)
+
 		job.add_stats(H)
 
 		job.add_knownCraftRecipes(H)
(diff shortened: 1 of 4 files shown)


Kazkin:
While interesting in theory, the issue is it doesn't solve what is meant to be addressed. The issue being that medical doesn't have tasks that they alone can do, medically speaking, if someone manages to get the right equipment or even a halfway decent oddity. I will archive this code, as the alternate titles giving alternate stats is a **very** interesting idea I will play with later.

PR #2819, opened 2022-03-03
RCDs

Lamasmaster:
## About The Pull Request
<details>
<summary>
	Firstly buffs existing RCD making it have double the capacity (granted it makes it eat entire stack of matter at once for only 60) while adding 2 new types (one of which is most likely adminspawn only) industrial (double capacity at 120, bit less costly on few designs and faster) and excelsior (double capacity of industrial and much faster and cheaper). At same time regular RCD can no longer deconstruct reinforced walls and only it's advanced versions can.
	Adds 4 more modes being grilles, windows, machine frames and computer frames which it can deconstruct now too. Window mode specifically has to be used on a floor under the low wall otherwise it ends up being in the air and shattering second you move it (tried to fix it but sadly this is only way it worked for me)
	Industrial RCDs can be purchased from lonestar for 4k and comes in package of 2, guild master starts with one in his secure storage instead of regular one.
	
RCD

</summary>
<hr>
	
<hr>
</details>

## Changelog
:cl:
add: New types of RCDs and functions to existing one
/:cl:


Kazkin:
Neat, good work.

PR #2840, opened 2022-03-06
ID cards get a title toggle verb.

LongSleeves:
Adds ability to toggle the title on ID cards.
No longer are well known people stuck having their first names removed with a worthless snowflake title.
Marshals and Blackshield can now go on break! (Then get yelled at by commander and warrent officer!)

Right clicking the ID card allows you to alter if it shows the title or not. Or you can use the "Toggle-Title" command for those who like to type. Tested with proof picture in development-general.


Kazkin:
Was this tested? Absolutely great idea but the return = "" worries me unless this was verified to work without runtimes.

Trilbyspaceclone:
image

PR #2851, opened 2022-03-07
Title Toggle HOTFIXED version~

LongSleeves:
moves the var (show_title) to the human.dm file. This means the card/id file is no longer modified by this PR.
Creates the verb Toggle Title for the IC tab and has it change on the person instead of card. This means dropping or losing your ID card wont cause the game to produce runtimes and start crying in the corner.


LongSleeves:
Should be fully functional now. With the var contained on the character instead of the card should be much better off for keeping everything close together too.

PR #2857, opened 2022-03-08
Cleans up back-end of hand-crafting

Risikio:
Adds a box of disks to the crafting menu.

Cleans up a lot of the back-end to hopefully make it easier to create crafting items based on perks.


Kazkin:
Need a review from both sieg and hex for this.

PR #2858, opened 2022-03-08
Adds in New Soteria Crafting

SouthyOsmanli:
## About The Pull Request
<details>
<summary>
	Adds in a few new things to Soteria. Firstly - new gauze and ointment, with 8 uses instead of 5. Slightly better over-all compared to their non-Soteria counterparts. Requires a special crystal made through chemistry. Also gave Soteria a set way of making more cloth instead of random 2-7 as normal, instead they can tear up any clothing to make a solid 4 due to the requirements of the gauze being 8. Sprites were all done by HOUND.
</summary>
<hr>
	
<hr>
</details>

## Changelog
:cl:
add: New Soteria crafting page.
add: New Soteria specific gauze and ointment.
add: New crystal-steralizing thingy used in crafting. And maybe later I'll make it smokable just so it can posion people for looking like a fancy crack-rock.
/:cl:


SouthyOsmanli:
Skill issue - don't get hurt and you wouldn't need meds. Smhing my head.

PR #2885, opened 2022-03-12
Psionic Powers : Heck Around And Find Out Edition.

r4d6:
This PR add multiple powers that @Kazkin asked me to code, with only one currently available to the public.
The rest will be added in a future update by Possum.

**Kinetic blast**: Standard power (all psions get it)
Creates a psi based item in your hand that fires a kinetic blast. The blast costs 1 psi point per shot. The item disappears if dropped. The damage scales off of cognition, dealing max damage if you have the psionic psychosis perk.

**Researched variants**
- **Cryo-Kinetic Blast**: 4 points per shot, deals no damage but stuns for 1 second per 10 points of cog, capping at 4.
- **Pyro-Kinetic Blast**: 2 points per shot, impacts like an explosive grenade, damaging objects, knocking people down, and setting flammable objects on fire.
- **Electro-Kinetic Blast**: Hit scan, costs 3 points per shot, acts like an upgraded kinetic blast that does more damage and hits harder while being instant.

Other Researched Powers
**Detect Thoughts** - Gives you the exact coordinates of a current person. They must be alive, organic, and cannot have a cruciform or be wearing a tinfoil hat. Costs 5 points.
**Psychoactive Manipulation** - Allows you to reroll an oddity's stats and perk, effectively making a new oddity in the hopes of getting a better outcome. 2 points.
**Kinetic Barrier** - Creates a semi-visible purple-ish barrier in front of the user that last 10 seconds. The barrier blocks all projectiles and movement but not vision. Cost 2 points. By default it spawn a 3x1 barrier. At 40+ cog, 5x1. The barrier is horizonal to the users current direction.
**Chosen Control** - Clears all pain, hallos, stuns, sleep effects, addictions, and removes handcuffs, buckling, and grapples. Clears the blood of all drugs, stims, and medicines, effectively giving you a full clean. Cost 4.

The Final Results may be different from the explications.

### DIFF
--- a/code/modules/organs/internal/psionic.dm
+++ b/code/modules/organs/internal/psionic.dm
@@ -44,7 +44,8 @@
 		/obj/item/organ/internal/psionic_tumor/proc/psychic_call,
 		/obj/item/organ/internal/psionic_tumor/proc/psychic_banish,
 		/obj/item/organ/internal/psionic_tumor/proc/journey_to_nowhere,
-		/obj/item/organ/internal/psionic_tumor/proc/psionic_armor
+		/obj/item/organ/internal/psionic_tumor/proc/psionic_armor,
+		/obj/item/organ/internal/psionic_tumor/proc/kinetic_blaster,
 	)
 
 /obj/item/organ/internal/psionic_tumor/psychiatrist
@@ -66,6 +67,7 @@
 		/obj/item/organ/internal/psionic_tumor/proc/psychic_banish,
 		/obj/item/organ/internal/psionic_tumor/proc/journey_to_nowhere,
 		/obj/item/organ/internal/psionic_tumor/proc/psionic_armor,
+		/obj/item/organ/internal/psionic_tumor/proc/kinetic_blaster,
 		// Psych unique powers just for them. Do not add these to other lists. -Kaz
 		/obj/item/organ/internal/psionic_tumor/proc/peace_of_mind,
 		/obj/item/organ/internal/psionic_tumor/proc/psionic_heal_other,
--- a/code/modules/organs/internal/psionic.dm
+++ b/code/modules/organs/internal/psionic.dm
@@ -45,14 +45,7 @@
 		/obj/item/organ/internal/psionic_tumor/proc/psychic_banish,
 		/obj/item/organ/internal/psionic_tumor/proc/journey_to_nowhere,
 		/obj/item/organ/internal/psionic_tumor/proc/psionic_armor,
-		/obj/item/organ/internal/psionic_tumor/proc/kinetic_blaster,
-		/obj/item/organ/internal/psionic_tumor/proc/cryo_kinetic_blaster,
-		/obj/item/organ/internal/psionic_tumor/proc/pyro_kinetic_blaster,
-		/obj/item/organ/internal/psionic_tumor/proc/electro_kinetic_blaster,
-		/obj/item/organ/internal/psionic_tumor/proc/kinetic_barrier,
-		/obj/item/organ/internal/psionic_tumor/proc/chosen_control,
-		/obj/item/organ/internal/psionic_tumor/proc/detect_thoughts,
-		/obj/item/organ/internal/psionic_tumor/proc/psychoactive_manipulation
+		/obj/item/organ/internal/psionic_tumor/proc/kinetic_blaster
 	)
 
 /obj/item/organ/internal/psionic_tumor/psychiatrist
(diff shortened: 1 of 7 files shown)


code/modules/organs/internal/psionic.dm
```diff
@@ -44,7 +44,15 @@
...
-		/obj/item/organ/internal/psionic_tumor/proc/psionic_armor
+		/obj/item/organ/internal/psionic_tumor/proc/psionic_armor,
+		/obj/item/organ/internal/psionic_tumor/proc/kinetic_blaster,
+		/obj/item/organ/internal/psionic_tumor/proc/cryo_kinetic_blaster,
+		/obj/item/organ/internal/psionic_tumor/proc/pyro_kinetic_blaster,
+		/obj/item/organ/internal/psionic_tumor/proc/electro_kinetic_blaster,
+		/obj/item/organ/internal/psionic_tumor/proc/kinetic_barrier,
+		/obj/item/organ/internal/psionic_tumor/proc/chosen_control,
+		/obj/item/organ/internal/psionic_tumor/proc/detect_thoughts,
+		/obj/item/organ/internal/psionic_tumor/proc/psychoactive_manipulation
```
hex37, line 55: Yo won't this give the new powers to all psions except psychiatrists?

code/modules/psionics/psionic_items/psi_weptools.dm
```diff
@@ -215,4 +215,114 @@
...
+	fire_sound_text = "laser blast"
+	max_upgrades = 0
+	var/projectile_type = /obj/item/projectile/kinetic_blast // What does it shoot
+	var/use_amount = 1 // How many psi-points is used per shot
+	var/mob/living/carbon/holder // The one that prevent the fist from fading
+	var/obj/item/organ/internal/psionic_tumor/PT // The psionic organ of the holder.
+
+/obj/item/gun/kinetic_blaster/New()
+	..()
+	START_PROCESSING(SSobj, src)
```
hex37, line 236: Where is the processing loop for this?

code/modules/psionics/researched_powers.dm
```diff
@@ -1 +1,145 @@
...
+
+/obj/item/organ/internal/psionic_tumor/proc/chosen_control()
+	set category = "Psionic powers"
+	set name = "Chosen Control (4)"
+	set desc = "Clears all pain, hallos, stuns, sleep effects, addictions, and removes handcuffs, buckling, and grapples. Clears the blood of all drugs, stims, and medicines, effectively giving you a full clean."
+	psi_point_cost = 4
+
+	if(pay_power_cost(psi_point_cost))
+		owner.visible_message(
+			"[owner] do something weird!",
```
hex37, line 79: ^does something wierd

code/modules/psionics/researched_powers.dm
```diff
@@ -1 +1,145 @@
...
+
+	var/obj/item/oddity/O = owner.get_active_hand()
+	if(!istype(O, /obj/item/oddity))
+		to_chat(owner, SPAN_NOTICE("This isn't the correct kind of oddity!"))
+		return FALSE
+	owner.visible_message("<b><font color='purple'>[owner] concentrates on the anomaly in their hand, something about it changing in a subtle way.</font><b>", "<b><font color='purple'>You focus on the energies around the object, swaying them to your will and enhancing it!</font><b>")
+
+	if(O.oddity_stats)
+		for(var/stat in O.oddity_stats)
+			O.oddity_stats[stat] = rand(1, O.oddity_stats[stat])
```
hex37, line 140: Doesn't this make it so the oddity's stats can only worse or, best case, just as good as the original?

code/modules/organs/internal/psionic.dm
r4d6: No, it is only for testing purposes

code/modules/psionics/researched_powers.dm
r4d6: I leave the emote to Possum

r4d6:
> Doesn't this make it so the oddity's stats can only worse or, best case, just as good as the original?
For some reasons I can't directly reply to it.
But it use the same thing that the original used to roll its original stats.

Yawet330:
This makes folken shaman entirely useless
-1 on the oddity reroll stuff
+1 for everything else

Yawet330:
Also, unless I'm drunk, all of these can be achieved by players? They've been added as psionic catalysts thus may(?) be able to be dropped by deepmaint entities

r4d6:
> This makes folken shaman entirely useless -1 on the oddity reroll stuff +1 for everything else

Nope. Because shamans have a +5 on whatever new stat they reroll, so they improve the oddity more, while the psy-power merely does a reroll like it was a new oddity.

> Also, unless I'm drunk, all of these can be achieved by players? They've been added as psionic catalysts thus may(?) be able to be dropped by deepmaint entities

The list of droppable catalysts is hardcoded, they will not be dropped.

PR #2888, opened 2022-03-12
Revert "quick fix for my wurm spawn increas"

Yawet330:
Reverts sojourn-13/sojourn-station#2887

This massively increased special chances too, which was undocumented. Also makes these burrows even more dangerous which as said in general and dev discussion, was already an issue as they spawn forever.


Yawet330:
closing for now, something funny came up and I'm not actually sure if burrows affect spawning at all lulw

PR #2889, opened 2022-03-12
Livestock and Biogen QoL changes

DerFlammenwerfer:
## About The Pull Request
<details>
<summary>
	Makes poultry much more accessible to both Garden and Lodge, gives them one more means of acquiring Diethylamine (more expensive than Robust Harvest), tweaks the biogen cost of salt and pepper to be in line with sugar, and slashes the kind of ridiculous egg box biogen cost from 6400 to a more reasonable but still steep 3200.
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
tweak: Reduces biogen price of salt and pepper to match sugar
add: adds diethylamine to the biogen for cost of 150 (twice the cost of Robust Harvest)
tweak: Increases hatchable egg chance for normal chickens to 1 in 4, to make poultry more accessible to the kitchen
tweak: Increases the max number of eggs per feeding a clucker may lay to match the chicken, and how many eggs it can accumulate before not being hungry.
balance: Makes the hatchable egg rate a 1-in-3 chance for clucker eggs to compensate for them now laying more per feeding.
balance: Reduces biogen cost of the box of eggs from 6400 to 3200; still rather expensive, and far less effective than just feeding hens, but at least now it's more of an option for upgraded ones.
fix: Cluckers now leave behind poultry instead of normal meat when butchered -- lodge can now eat mor chikin

/:cl:

### DIFF
--- a/code/game/machinery/biogenerator.dm
+++ b/code/game/machinery/biogenerator.dm
@@ -20,18 +20,19 @@
 		"Food",
 			list(name="Milk, 30u", cost=60, reagent="milk"),
 			list(name="Soy Milk, 30u", cost=60, reagent="soymilk"),
-			list(name="Black Pepper, 30u", cost=60, reagent="blackpepper"),
+			list(name="Black Pepper, 30u", cost=30, reagent="blackpepper"), //Cost reduced to be in line with Sugar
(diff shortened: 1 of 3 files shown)


Kazkin:
Seems reasonable, hopefully people dont creature spam.

PR #2902, opened 2022-03-14
Add a psionic effect to the artifacts

r4d6:
## About The Pull Request

This PR add to every procedural artifact that Xenoarch can dig up a chance (currently 100% for testing purposes) to have a psionic component that only psions are able to detect.

There is also a new machine that is able to take those artifacts and convert them, after a bit of wait, into psionic catalysts that everyone can use.

While it is currently hijacking the Artifact System, due to it already having a system for artifact distribution, it can easily be converted to turn another type of artifact in case Possum want to turn it into its own system.


Kazkin:
Conflicts, plus I'm gonna button this for another day where I can do some extensive coding.

PR #2921, opened 2022-03-18
Revert "tg sprites are shockingly good"

Yawet330:
Reverts sojourn-13/sojourn-station#2911

They were shockingly bad. Toolbox looks like a magic lunchbox, and the defibs make my eyes bleed :(

On a more serious, and condensed note, many of the added sprites while good for /tg/, do not look good at /all/ here; example one being the toolboxes - the second example being the new PACMAN/etc sprites; which were directly taken from /tg/ and stand out like an eyesore.

That being said, some of them ARE good, hence why I'm asking this to be turned from a direct revert into a selective one.


Yawet330:
image
image

Yawet330:
In an ideal world we'd selectively revert what changes we wanted, because uh.

E w

Some of the sprites are good.

PR #2950, opened 2022-03-22
Add Ameridian Crystals

r4d6:
## About The Pull Request
Add ~~Tiberium~~ **Ameridian** Crystals, a radioactive crystalline structure new to Amethyn that grow and spread on its own, with crystals that can be excavated for both financial and material gains.
But if enough crystals are packed together, one's harvesting operations might find itself being interrupted by the sudden appearance of ameridian golems.
Harvested crystals can be directly sold, or inserted into a Ameridian Processor to be converted into any non-organic material.
There is also a Ameridian Spire, a bigger, purplier crystal that is practically indestructible and constantly spawn crystals around it.

Also steal Eris' golem sprites & add a new mutation that make the user immune to radiation damage.


r4d6:
The main thing that need to happen is to balance the inputs and outputs of the Tiberium Processor.

DerFlammenwerfer:
The cities of Nod will rise!

Trilbyspaceclone (changes requested):
Credit the artest that made the golems

Trilbyspaceclone (approved):

PR #2964, opened 2022-03-25
Massive Slime Waifu Rework

r4d6:
Directives from Possum : 
> Slime People
> Automatic effects
> -Can reach into bags without taking them off, including duffels.
> -Cleans floors and gains nutrition for doing so.
> -Takes 15% more damage from all sources.
> -Immunity to oxygen damage, toxins, and infections.
> -When taking toxin damage, is instead healed.
> -Only organ is the brain.
> -Injecting liquid plasma into their slime core after dying regrows their body.
> -Gibs on death, leaving behind the core.
> -Ethnicity types give them slime-based implants.
> -Will never be attacked by regular slimes under any circumstance, even when they turn hostile to all other types.
> 
> Perks
> -Can regrow limbs.
> -Self inject a non-od variant of universal healing chem.
> -Enhance mental stats.
> -Enhance physical stats..
> -Can boost speed.


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -48,7 +48,36 @@
 		if(germ_level < GERM_LEVEL_MOVE_CAP && prob(8))
 			germ_level++
 
-
+		if(src.species?.reagent_tag == IS_SLIME) // Slimes clean the floor. Code stolen from the roombas
+			var/turf/tile = loc
+			var/nutrition_gained = 5 // Up here for ease of modification
+			if(isturf(tile))
+				tile.clean_blood()
+				for(var/A in tile)
```
hex37, line 56: this for loop is a little inefficient for something that will run every single time a slime moves, it could make a lag machine pretty easily

Maybe setting up a signal proc would be better? So here it would send

SEND_SIGNAL(tile, COMSIG_CLEAN, src)

And that would link up to procs defined by using RegisterSignal for functions objects, effects and creatures within that tile.

r4d6: 1. I have no idea how the heck signals works.
2. It is the same code that the roombas use to clean the tile.

r4d6:
I did some minor testing, and the MC isn't going to break with this PR, so it is fine to be merged, even as incomplete as it is.

PR #2990, opened 2022-03-27
Corrects long-standing bug with crafting the nanoforge

Yawet330:
Legacy code caused this machine to need two boxes when being built, but only returned one.

Was because we made it able to take new artifacts on the fly, without having to rebuild it.


Yawet330:
Turns out, this is intended, and was actually the result of it needing *a* artifact, just not the box to operate.

Catalyst power is apparently done via checking the current strength + the perk, so technically any perked mechanical oddity can be used in it, I'm fine with that then lmao

PR #3010, opened 2022-04-02
crossbow buff

lolman360:
## About The Pull Request
buffs crossbows so that the string can be drawn back while moving

currently it takes 10 seconds of standing still to get a crossbow to full tension, this at least allows you to move around while yo do it
	
## Changelog
:cl:
balance: crossbow/rxd buff
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/launcher/crossbow.dm
+++ b/code/modules/projectiles/guns/launcher/crossbow.dm
@@ -65,6 +65,7 @@
 	cell = null    // Used for firing superheated rods.
 	var/current_user                        // Used to check if the crossbow has changed hands since being drawn.
 	var/draw_time = 20							// How long it takes to increase the draw on the bow by one "tension"
+	var/stationary_draw = FALSE
 
 /obj/item/gun/launcher/crossbow/update_release_force()
 	release_force = tension*release_speed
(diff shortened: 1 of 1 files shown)


Guidesu:
This crossbow are bay's legacy code.
I was just going to do an PR to re-work them into an real gun.
Its not worthy it

lolman360:
their code is fine tbh. and they'd be fine balance-wise with this buff

Kazkin:
I could have swore I removed this entirely.

ThePainkiller:
A great step on the right direction, if only their (non Lodge version's) damage wasn't crap...

lolman360:
> A great step on the right direction, if only their (non Lodge version's) damage wasn't crap...

i actually really like the metalrod throwing crossbows, they actually feel better than lodge ones to use. i might consider buffing damage to bring them up to lodge spec, considering they still take 10 seconds to charge a full-power shot

ThePainkiller:
> > A great step on the right direction, if only their (non Lodge version's) damage wasn't crap...
> 
> i actually really like the metalrod throwing crossbows, they actually feel better than lodge ones to use. i might consider buffing damage to bring them up to lodge spec, considering they still take 10 seconds to charge a full-power shot

I've tried it countless times as outsider when spawning on the woods (and couldn't get the rifle due to mukwahs) and sincerely prefer kiting on melee than waiting ~10 seconds to load a round that SOMETIMES EVEN MISSES ITS TARGET and causes about less damage than a .35 fired from a Havelock

lolman360:
i just checked the code and unless i'm missing something crossbows don't multiply throwforce and are launchers, meaning the max damage one can do is 15

testing it now because if so i'll probably rework them so they do useable damage

ThePainkiller:
> testing it now because if so i'll probably rework them so they do useable damage

Gods fucking bless you lolman. There was literally no difference on making kitchen knives and throwing them normally than with using these, as "throwers" their projectiles can literally miss unlike proper "guns"

lolman360:
image
@ThePainkiller

lolman360:
closing, will refactor crossbows soon

PR #3013, opened 2022-04-02
[REVERT] Reverts trilby's nerf to funny man viro

Yawet330:
This made longevity so trash it wasn't even worth using. No.

To explain why, viro symptoms DO NOT HAVE A 100% chance to proc.

This was not tested in its pre-nerf state, this was not tested in its after-nerf state. As the only individual who has tested this, it was fine. It healed ~30 damage of brute in 5 minutes.

### DIFF
--- a/code/modules/virus2/effect.dm
+++ b/code/modules/virus2/effect.dm
@@ -166,9 +166,9 @@
 		if(ishuman(mob))
 			var/mob/living/carbon/human/H = mob
 			for (var/obj/item/organ/external/E in H.organs)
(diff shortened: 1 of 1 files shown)


Yawet330:
For reference: During the assault on the excelsior base, I gave Thatcher and a few others the disease and let it mature.

It did not even manage to out-heal space damage, with a base healing of 5 at the time.

cdb-is-not-good:
image

Dongels:
:salt:

PR #3015, opened 2022-04-03
'Black Arrow' railgun for guild, minor guild crafting buff

lolman360:
## About The Pull Request
New "Black Arrow" railgun. Takes metal rods and a large cell, fires low- to hyper-velocity projectiles. High firedelay, very bad one-hand penalty, somewhat low recoil.

Generally worse than the reductor, given its higher cost (needs uranium), higher firing cost (secondary ammunition type), slowdown on hold and slower firerate. 

values probably subject to change

Guild recipe stat COG >> MEC. Makes more sense, as Guild typically has highest MEC, and they do a lot of crafting. This is just a straight Guild buff, though.

## Changelog
:cl:
add: "Black Arrow" railgun for Guild. Fires metal rods.
balance: Guild primary crafting stat is now MEC, not COG.
/:cl:


Yawet330:
Why would you ever build this? It's a direct downgrade from the reductor and costs more.

lolman360:
> Why would you ever build this? It's a direct downgrade from the reductor and costs more.

1. reductor is arguably overpowered in its current state

2. higher AP, higher base damage (better with multipliers), can be suppressed, integrated scope with better zoom

Kazkin:
Right, sorry for the delay on the review of this PR.

A few things:
-75 damage is higher than the reductor.
-2.0 scope is a further distance than even the anti-material rifle.
-The sprites do not meet our spriting standards.

The reductor has the same damage as a standard shotgun, with its lethal rail shot being 1:1 to a shotgun slug besides not having as much damage fall off and being hit scan. The subtype of the reductor, the myr, is 1:1 with kurtz .50. The damage of 75 makes it arguably better than .60-06 ammo, the same as anti-material rounds.

lolman360:
> Right, sorry for the delay on the review of this PR.
> 
> A few things: -75 damage is higher than the reductor. -2.0 scope is a further distance than even the anti-material rifle. -The sprites do not meet our spriting standards.
> 
> The reductor has the same damage as a standard shotgun, with its lethal rail shot being 1:1 to a shotgun slug besides not having as much damage fall off and being hit scan. The subtype of the reductor, the myr, is 1:1 with kurtz .50. The damage of 75 makes it arguably better than .60-06 ammo, the same as anti-material rounds.

damage is down to 60/60/70,  fire delay is up to 16/18/20 on lv/hv/uhv. ranges have been adjusted, scope is now slightly better than the watchman (1.4 vs watchman's 1.2). 2.0 is the same as an AMR, not more.

max effective range has been reduced to 18 from 30, and suppressed effective range is down from 10 to 7.


as for the sprite, should I just change the palette to one of our currently used ones?

Kazkin:
The damage is still absurdly high but I can balance that myself in time, as for the sprite, its major issue is the stock and barrel. In general, you should *always* avoid using 1 pixel for any object, in particular objects that connect other objects, the barrel is weirdly thin, the stock is oddly connected, and the gray scale tone is rather ugly. It needs better coloring/shading.

lolman360:
> The damage is still absurdly high

Damage is actually *less* than the reductor at most levels, since it has 64.8 damage firing slugs (it has 1.2x damagemod)

I'll do some more work on the sprite, though.

Kazkin:
Better but not quite there. The shading is good but the barrel definitely needs work.

lolman360:
niche filled by batko from #3056

PR #3017, opened 2022-04-03
Lonestar Organ Market

Jul-Jul:
## About The Pull Request
<details>
<summary>
	Simply allows Supply consoles to buy contraband, but more importantly sell contraband without the need of a EMAG. In this case, allowing the exporting of organs and other questionable items.

Considering we don't have any contraband that we can buy, and getting hold of a EMAG is near impossible. I believe this PR simply allows Lonestar to export more...questionable things. 
</summary>
<hr>	
<hr>
</details>

## Changelog
:cl:
Edit Supply console.
/:cl:

### DIFF
--- a/code/game/machinery/computer/supply.dm
+++ b/code/game/machinery/computer/supply.dm
@@ -12,10 +12,10 @@
 	var/temp
 	var/reqtime = 0 //Cooldown for requisitions - Quarxink
 	var/last_viewed_group = "categories"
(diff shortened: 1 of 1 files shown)


Kazkin:
Reasonable, but if the emaggs get abused even once I'm removing them or increasing the price to 50k per emag.

Jul-Jul:
I don't believe EMAGs can even be bought even with this, correct me if I'm wrong.

PR #3020, opened 2022-04-04
Fix Unknown speech

MSieghardt:
this will fix the issue of people appearing as 'unknown' despite wearing a mask. now they will be "Unknown as" since their id says who they are.

tested it and it works.

### DIFF
--- a/code/modules/mob/living/carbon/human/say.dm
+++ b/code/modules/mob/living/carbon/human/say.dm
@@ -137,6 +137,9 @@
 			var/obj/item/clothing/mask/mask = wear_mask
 			if(istype(mask) && mask.muffle_voice)
 				voice_sub = "Unknown"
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/say.dm
```diff
@@ -137,6 +137,9 @@
 			var/obj/item/clothing/mask/mask = wear_mask
 			if(istype(mask) && mask.muffle_voice)
 				voice_sub = "Unknown"
+				if(GetIdCard())
+					var/obj/item/card/id/gotcard = src.GetIdCard()
+					voice_sub = voice_sub + " as [gotcard.registered_name]"
```
r4d6, line 142:
```suggestion
					voice_sub += " as [gotcard.registered_name]"
```

PR #3023, opened 2022-04-04
Nanogate nerds take more cloneloss

hex37:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
Nanogate users now actually take cloneloss damage when mutations are injected in them- and a good bit more of it.
/:cl:

### DIFF
--- a/code/modules/genetics/creatures/wasonce.dm
+++ b/code/modules/genetics/creatures/wasonce.dm
@@ -151,10 +151,10 @@ Has ability of every roach.
 				to_chat(captive, SPAN_DANGER(pick("The immense strength of the creature is crushing. Wasn't... Flesh supposed to be weak?")))
 				captive.adjustBruteLossByPart(15, pick(captive.organs))
 			else
-				injector.inject_mutations(captive, TRUE)
 				to_chat(captive, SPAN_DANGER(pick(
 					"The mass changes you...", "Veins slip into your flesh and merge with your own", "Parts of yourself fuse to the roiling flesh surrounding you.",
 					"You feel yourself breathing through multiple lungs.", "You feel yourself assimilating with the whole.")))
+				injector.inject_mutations(captive, TRUE)
 		injector.removeAllMutations()
 	if(lethal_to_captive && captives.len && prob(15))
 		for(var/mob/living/captive in captives)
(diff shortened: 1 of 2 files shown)


Yawet330:
Massive no. This isn't a bug fix, this is a balance change which makes 10 mutations do 100 clone damage.

This did not need a buff towards its damage, but merely be fixed.

r4d6:
> Massive no. This isn't just a bug fix, this is a balance change which makes 10 mutations do 100 clone damage.
> 
> This did not need a buff towards its damage, but merely be fixed.
> 
> (making it do 5x its original damage is kinda :^(, especially when nanogate users already take additional damage from being inside the blob in the first place, like synths)

Assuming that every players is in otherwise perfect health and has 100 HP, it would take 50 mutations for them to die from it with the old damage.
With the new damage, they need 10 mutation total.

If what I am being told is true, the blob inject a mutation every 5 seconds the user spend inside it. Meaning that at the current damage level, it would take 50 seconds before the user die of clone damage, not counting time dilatation which would make it be longer.
Previously, they would be able to spend 250 seconds before death, assuming they got damaged correctly. Or 4 whole-ass minutes.

Yawet330:
> > Massive no. This isn't just a bug fix, this is a balance change which makes 10 mutations do 100 clone damage.
> > This did not need a buff towards its damage, but merely be fixed.
> > (making it do 5x its original damage is kinda :^(, especially when nanogate users already take additional damage from being inside the blob in the first place, like synths)
> 
> Assuming that every players is in otherwise perfect health and has 100 HP, it would take 50 mutations for them to die from it with the old damage. With the new damage, they need 10 mutation total.
> 
> If what I am being told is true, the blob inject a mutation every 5 seconds the user spend inside it. Meaning that at the current damage level, it would take 50 seconds before the user die of clone damage, not counting time dilatation which would make it be longer. Previously, they would be able to spend 250 seconds before death, assuming they got damaged correctly. Or 4 whole-ass minutes.

You're disregarding the fact they ALSO take 10 combined damage from 5 brute/burn a tick, which means they take a total of 20dps, 10 of which can NEVER be healed and both of which ENTIRELY bypass armor.

This is, again, a stupidly strong buff to the jellies. Make them do more brute/burn, don't make them do more incurable damage.

PR #3051, opened 2022-04-08
Raise the maximum chicken limit by 10, to 20

r4d6:
Raise the universal maximum chicken limit by 10, to a limit 20
5 chicken, or half the current limit, were in the Admin Genetic Testing area, thus just about blocking more chicken from spawning from the eggs.
It was either raising the limit, or doing map work to remove the chicken crate.


ThePainkiller:
Another solution would've been to just delete the chickens there entirely? Since it's an admin test area, why not spawn the chickens themselves to do it?

r4d6:
> Another solution would've been to just delete the chickens there entirely? Since it's an admin test area, why not spawn the chickens themselves to do it?

It was this PR or mapping the crate out. I just chose the easiest solution.

PR #3072, opened 2022-04-12
Map in an ameridian crystal field

r4d6:
## About The Pull Request
This PR is about adding a Lonestar-seeded crystal field in the unused hole that was in the middle of mining.
- The Sonic Fence keep everything in, including the golems, and start the round active and powered.
- All the crystals start at the first level of growth.
- There is also a bottle of liquid ameridian for emergency reseeding.
- There is a Rad Suit in a Suit Storage Unit to protect the one harvesting the field from radiation.
- There is an handheld sonic emitter for whoever is in charge of harvesting the field to protect themselves.
- The fence + the walls mean that you can only be irradiated when in the hole, unless the sonic fence fail and let the crystals grow, at which point they would still be relatively contained by the mining base itself, preventing it from going out of proportion unless someone tease the golems and make them destroy the walls, at which point it's on you.
- Replace the R-Glass around the crystal field with R-Plasma Glass
- Replace the smol-ass fuel tank with the bigger variant
- Also add an Ameridian Refinery & a Liquid Ameridian Processor in the same room as the regular smelter.

Overall these changes should help Miners have an alternative to the Termite Drill and the regular smelting, while hopefully not being overpowered.


r4d6:
Approved by Lich
Approval

PR #3077, opened 2022-04-13
A shield nerf, gasp

cdb-is-not-good:
removes the Watchtower, Rushing Bull, and Lamia from the shield starting loadout - they were simply too good to start with and often overshadowed other choices, go get them from the weapon disk.
Slightly buffs the watchtower to +.3 scoped damage and default pen from .8 to 1.0(subject to change.)
Triage now starts with 5 mags instead of 3, same as all other shield primaries.
Triage damage mod upped by .15 to .75, this should help it feel a bit less anemic compared to other options with larger magazines and less expensive ammo
Nerfs the sky :pensive: now uses proper firemodes that respect existent rules about damage reduction. Also gets a 2 shot burst instead of three.
Firefly marshal sidearm kit replaced with counselor sidearm
Buckshot SLIGHTLY buffed. to compensate for my previous removal of AP it has been given an equivalent amount of halloss per pellet(5). This should make it a bit more useful in PVE
Gorget no longer covers ears and should be a little less janky
Fixes the wrestling belt, now has an on-mob sprite
Marshals and Blackshield no longer start with their vouchers on their person, they are now in their locker. This was a needed given how disposable these starting weapons were becoming, especially on high-pop rounds.
claimed lockers will now self-tag with (USED) when reset, allowing you to tell when someone has previously claimed a locker.
Can now crush dead simple/superior mobs with harm intent

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


lolman360:
remove warthog and duty (para is probably okay i havent seen it) and maybe bs will finally have balanced vouchers

cdb-is-not-good:
the warthog/duty are Sergeant only - though having them as a 'dispensable' option is a bit ??? imho, but it never comes up. It's cheaper to just buy materials and make the gun than to buy it in the vendor

PR #3079, opened 2022-04-13
Blackshield Adjustments

Noightmare:
Adjusts the Shield teleporter's energy gain slightly and its rewards for doing the acquirement quests, incentives shields to actually get said items. 
Adds in several new items that can be acquired from the teleporter, ranging from armor, ammo, to shield gun disk. Reduced point cost for plasteel, bringing it down from 250 to 150. 
Reduces license point cost of the Duty on the Blackshield disk. It was too much for it to be 6 license points for what it was worth.
Moved the ammo disk back into the secure armory. Was a mistake to have it outside.
Readjusts Blackshield ablative armor, giving it slowdown and raises some more general protections against other types, but better to wear other armor for said stuff.
Adds in more traps for those who tunnel in farther into the colony.

### DIFF
--- a/code/datums/contract.dm
+++ b/code/datums/contract.dm
@@ -585,7 +585,7 @@ GLOBAL_LIST_INIT(blackshield_item_targets,list(
 //TODO: Make let them send in other things/place items in areas to allow for addionation points
 /datum/antag_contract/blackshield/appropriate
 	name = "Appropriate"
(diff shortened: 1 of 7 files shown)


Kazkin:
Plasteel changed back to 250, its already too easy to get. Buy from other factions if you can't afford the 250.

PR #3100, opened 2022-04-15
Church Garden remodel

Spoopy11:
Church garden somewhat reworked:
 - Bees, ALL THE BEEEEEEEEEEEEEEEES
 - Put a fence around the garden and the solars to assist in guiding people around
 - Added COWS and changed the ugly walls into fenceses
 - A few bits repositioned to make it looks nicer

image
image

### DIFF
--- a/maps/__Nadezhda/map/_Nadezhda_Colony_Omni_STU.dmm
+++ b/maps/__Nadezhda/map/_Nadezhda_Colony_Omni_STU.dmm
@@ -3937,6 +3937,13 @@
 	name = "murky water"
 	},
 /area/nadezhda/maintenance/undergroundfloor1north)
+"aSA" = (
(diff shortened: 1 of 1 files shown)


Kazkin:
This has to be approved by lich but can't say I'm a fan. Chickens can ignore railings and escape and the church can produce cow milk through the bioprinter, removing the need to own cows.

Spoopy11:
> This has to be approved by lich but can't say I'm a fan. Chickens can ignore railings and escape and the church can produce cow milk through the bioprinter, removing the need to own cows.

Ah that's fair.  I was talking with a couple other church players, and they were discussing putting a biogenerator in the area to negate the need for chickens.  I wonder if we should plain add eggs to the disk and remove the animals in general/replace them with some sort of mascot animal

PR #3107, opened 2022-04-16
Minor addition to cargo packs

Chevalt:
This change adds new 7.5 and 257 ammo packs, aswell as orders for ammo disks and a last void wolf "Marksman" kit with a single bolt gun, a few rounds, and a mod. All of these were designed to be accurately priced so they're not overpowered, but to also make them relatively accessible for departments when cargo is /actually/ there.

I was tired that there was only handgun ammunition in the cargo console so I added more options.


Chevalt:
second commit is a minor fix, removing the "" that made some checks run false.

PR #3113, opened 2022-04-16
temperature tolerance changes

cdb-is-not-good:
Important note; These values are all in kelvin. the base values are inherited directly from Eris and thus is likely why they are so "peculiar" and wide in range(and they were the base for our alien races and subsequently influenced said races values.)

Large changes to temperature tolerance ranges for all races. Initial values saw most races essentially immune to the affects of temperature outside of very specific scenarios.  To give some more context, tolerable temp is any value lower(or higher) than the 'cold/hot level 1-2-3'. This means that, say - for humans, whos heat-level-1=360, they did not suffer ANY ill effects of temperature until 360k(this is about 188f or 86c) and this is why we had weirdness with say...people getting infections and not even noticing till they were already acute, because they didn't even hit the warning stage of temperature before entering acute. New values are intended to be a bit less "silly" without being overly realistic or punishing.

BASE levels(includes human, exalt, etc. Any 'unchanged' races that use default values use these numbers)
cold_level_1 =260->270
cold_level_2 =200->230
cold_level_3 =120->200
heat_level_1 =360->330
heat_level_2 =400->380
heat_level_3 =1000->460

Sablekyne(Kept their broad range of temperature resistance while given some sanity in numbers.)
cold_level_1 =200->240
cold_level_2 =140->215
cold_level_3 =80->190
heat_level_1 =420->340
heat_level_2 =460->400
heat_level_3 =1200->480

Mar'Qua
cold_level_1 =140->230
cold_level_2 =80->210
cold_level_3 =20->190

Akula
cold_level_1 =200->240
cold_level_2 =140->200
cold_level_3 =80->170

Cindarite
cold_level_1 =280->290
cold_level_2 =220->265
cold_level_3 =140->240
heat_level_1 =450->370
heat_level_2 =520->410
heat_level_3 =1400->500

Additionally, nerfs the hell out of hot-ramen to accompany this change - neverminding the new temperature ranges, it should be notably less deadly now.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species.dm
+++ b/code/modules/mob/living/carbon/human/species/species.dm
@@ -68,12 +68,12 @@
 	var/breath_type = "oxygen"                        // Non-oxygen gas breathed, if any.
 	var/poison_type = "plasma"                        // Poisonous air.
 	var/exhale_type = "carbon_dioxide"                // Exhaled gas type.
-	var/cold_level_1 = 260                            // Cold damage level 1 below this point.
-	var/cold_level_2 = 200                            // Cold damage level 2 below this point.
-	var/cold_level_3 = 120                            // Cold damage level 3 below this point.
-	var/heat_level_1 = 360                            // Heat damage level 1 above this point.
-	var/heat_level_2 = 400                            // Heat damage level 2 above this point.
-	var/heat_level_3 = 1000                           // Heat damage level 3 above this point.
+	var/cold_level_1 = 270                            // Cold damage level 1 below this point.
+	var/cold_level_2 = 230                            // Cold damage level 2 below this point.
+	var/cold_level_3 = 200                            // Cold damage level 3 below this point.
+	var/heat_level_1 = 330                            // Heat damage level 1 above this point.
+	var/heat_level_2 = 380                            // Heat damage level 2 above this point.
+	var/heat_level_3 = 460                           // Heat damage level 3 above this point.
 	var/passive_temp_gain = 0		                  // Species will gain this much temperature every second
 	var/hazard_high_pressure = HAZARD_HIGH_PRESSURE   // Dangerously high pressure.
 	var/warning_high_pressure = WARNING_HIGH_PRESSURE // High pressure warning.
@@ -93,8 +93,8 @@
 
 	var/list/perks = list()
 
-	var/heat_discomfort_level = 315                   // Aesthetic messages about feeling warm.
-	var/cold_discomfort_level = 285                   // Aesthetic messages about feeling chilly.
+	var/heat_discomfort_level = 330                   // Aesthetic messages about feeling warm.
+	var/cold_discomfort_level = 270                   // Aesthetic messages about feeling chilly.
 	var/list/heat_discomfort_strings = list(
 		"You feel sweat drip down your neck.",
 		"You feel uncomfortably warm.",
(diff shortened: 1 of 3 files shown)


ThePainkiller:
While it has always been hilarious to see someone with a _"fever"_ of around fucking **86º celcius** (Barely 14 degrees beneath the boiling point of water, or 360 kelvin in burger units so you get the idea) not spontaneously combusting or passing out, and not even **NOTICING** they have a fever (because nobody pays attention to the temperature meter), this change makes it more realistic in approach BUT I have a couple questions regarding balance and possible issues this might cause:

- Have you taken into consideration Frost Oil's temperature adjustment? I see it has been untouched completely and with these changes it has become even more lethal than it was (heavy slowdown through cold is already dangerous and annoying enough, top it off with general bodypart damage and Cave Spiders are now far, far more lethal)
- Have you taken into consideration the temperature changes caused by fire stacks? I fear this will soft-roll back the agony changes to fire stacks that were introduced not long ago to make fire stop being absurdly lethal (even moreso wearing SCAF suits or any voidsuit for that matter), this definitely will make getting hit by a Greyson cook robot a guaranteed death due to pain-through-fire getting easier to stack.
- Have you considered adding `span` notices like toxins damage do with **"Your whole body hurts a little."** whenever your temperature reaches a certain threshold? Something like _"Your body feels sweaty/considerably hot/cold."_? I feel like this would be another way that would make characters realize they **do** have a fever and something's wrong with their bodies in a way that the temperature meter fails to accomplish, rather than bumping numbers up and down for the sake of actually receiving damage so that they realize their infections?

cdb-is-not-good:
>Have you taken into consideration Frost Oil's temperature adjustment? I see it has been untouched completely and with these changes it has become even more lethal than it was (heavy slowdown through cold is already dangerous and annoying enough, top it off with general bodypart damage and Cave Spiders are now far, far more lethal)
Yes, I have taken it into consideration and looked at it here 'n there. It's part of why say - hot ramen was reduced. I had also planned to make Capsaicin less deadly but it turned out to not need it. It's something that WILL need testing and i'll likely have to look at frost-oil to make sure it doesn't increment your temperature too much per tick - though in general, I did leave room for some of these things with the numbers.

>Have you taken into consideration the temperature changes caused by fire stacks? I fear this will soft-roll back the agony changes to fire stacks that were introduced not long ago to make fire stop being absurdly lethal (even moreso wearing SCAF suits or any voidsuit for that matter), this definitely will make getting hit by a Greyson cook robot a guaranteed death due to pain-through-fire getting easier to stack.
Also, yes. fire will likely be deadlier not due to pain(as temperature=/=agony), but due to the lower threshhold at which you take different levels of damage. That said, we have fire-immune suits, and more suits offer at least partial resistance to fire than previously.

>Have you considered adding span notices like toxins damage do with "Your whole body hurts a little." whenever your temperature reaches a certain threshold? Something like "Your body feels sweaty/considerably hot/cold."? I feel like this would be another way that would make characters realize they do have a fever and something's wrong with their bodies in a way that the temperature meter fails to accomplish, rather than bumping numbers up and down for the sake of actually receiving damage so that they realize their infections?
That's not the main reason for this PR. The main reason for this PR is that the numbers are frankly a little ridiculous, and while i don't exactly want to 'realify' things by making the numbers completely punishing, i did want to see a change to them to make things a bit more "sane"? span messages would be nice - such messages DO Exist for temperature but for SOME reason only play when the room is that temperature, as opposed to your body.

PR #3118, opened 2022-04-17
GetFlatIcon() optimizations from 2018 that we dont have yet

lolman360:
god why
untested


MSieghardt:
Things look fine, nothing really breaks. Not feeling too well so I didn't analyze how it'd improve stuff too much, but at least it doesn't seem to break it.

PR #3134, opened 2022-04-18
Removes the forced harddel on mobs

nikothedude:
## About The Pull Request
A while ago, eris updated their mob file to force harddels on everything. THis was due to a very old TG PR. TG has moved on from that as it was a temporary, awful measure. See https://github.com/tgstation/tgstation/pull/51887.

WARNING. I just found out this causes most mobs to fail GC on deletion. This means they are stuck in the memory and can cause a memory leak. This isnt to say its our only memory leak, we have a fuckton of GC failures, this'll jsut make it worse.

	
## Changelog
:cl:
code: Updated mob destroy() to not force harddels.
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -11,8 +11,7 @@
 		client.screen = list()
 
 	ghostize()
-	..()
-	return QDEL_HINT_HARDDEL
+	return ..()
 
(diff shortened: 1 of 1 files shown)


nikothedude:
Do note this might cause issues with blob, however, we don't use blob much, and it's bound to reduce lag.

keep an eye on hivemind, though.

PR #3145, opened 2022-04-19
Material additions to cargo-engineering

Chevalt:
More materials options have been added to the engineering section of cargo, they're quite heavily priced, as to make miners the more viable choice for when they're actually there, but to also make materials-heavy departments like engineering and soteria viable when miners aren't actually there.

### DIFF
--- a/code/modules/cargo/packs_engineering.dm
+++ b/code/modules/cargo/packs_engineering.dm
@@ -167,6 +167,64 @@
 	crate_name = "plasteel sheets crate"
 	group = "Engineering"
 
+/datum/supply_pack/diamond15
+	name = "Diamond Sheets Crate (15)"
+	contains = list(/obj/item/stack/material/diamond)
+	amount = 15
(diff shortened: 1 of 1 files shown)


cdb-is-not-good:
Pretty good, a bit painful but that's okay. We don't want to encourage people to be buying here instead of via miners/scavs. Convenient.

Kazkin:
This needs a price comparison against the trade beacon ship that already supplies ores/sheets. The price should be == or greater, depending on how we want to do this. I'm actually against this PR though, as the shuttle trade beacon is already a way to get overpriced materials.

Kazkin:
I'll leave this up until someone does the price comparisons.

ThePainkiller:
Oh, I thought this added the rare material stuff like osmium and the like, it's not that bad then

BigBadPossum:
With no comparison forthcoming, we'll go ahead and close this.

PR #3150, opened 2022-04-19
SMG balance stuff

cdb-is-not-good:
An simple change to SMG weapons(and only SMG weapons). Currently, they're in a very weird place where a generally low damage mod, plus pistol caliber(.35 almost exclusively) and further penalties for using full auto meant that they were highly unfavorable to use in almost all scenarios. Consequently, adds new fire modes without the -.2 damage mod ONLY for SMGs. This will hopefully help their lackluster performance in almost all fields when not being used essentially as large pistols, which you'll note retain their damage penalty for full auto.  

This change affects the c20r, Drozd, Greasegun, Vintovka Lyugera, Mac10, PPSH, Texan, Ekaterina, Semyonovich, Thompson

Looking at .35 and .40 in a vacuum, this will amount to a total damage bonus of 4.8 per shot on .40 SMGs(24->28.8) and a total damage bonus of 3.2(16->19.3) on .35. Obviously, these numbers may need some tweaking but overall should make SMGs less of a collection of depressing, trash guns.

Additionally, SMGs may no longer be braced by default(this is subject to change, we'll see.), without going and getting a proper bipod to use.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/__DEFINES/weapons.dm
+++ b/code/__DEFINES/weapons.dm
@@ -52,12 +52,26 @@
 #define FULL_AUTO_600		list(mode_name = "full auto",    mode_desc = "600 rounds per minute",   mode_type = /datum/firemode/automatic, fire_delay = 2  , icon="auto", damage_mult_add = -0.2)
 #define FULL_AUTO_800		list(mode_name = "fuller auto",  mode_desc = "800 rounds per minute",   mode_type = /datum/firemode/automatic, fire_delay = 1,   icon="auto", damage_mult_add = -0.2)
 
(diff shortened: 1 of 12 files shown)


Spoopy11:
Worth making pistols not braced by default or out of scope?

cdb-is-not-good:
I did actually think about that, yea. Just didnt wanna go to the trouble of doing that for every pistol RN since this is just a little balance PR and honestly, i dont see it come up too often for pistols.

Kazkin:
SMG's are very rarely used and this PR pretty much addresses why, even if the dps is high, few people want to expend more bullets at a huge damage loss. Approved.

PR #3162, opened 2022-04-20
Xenobio Dye Vat -Admin Testing Only For Now-

LongSleeves:
Base functionality added to the slime_dye_vat (Has to be spawned in)
Uses crayons to set color, Slime jars to get uses for dying things. And of course.... Comes with its own special creation!
Currently the base model is uncolored. And there is two known colors that don't appear right. Orange and Rainbow.

### DIFF
--- a/code/game/machinery/xenobio.dm
+++ b/code/game/machinery/xenobio.dm
@@ -68,7 +68,7 @@
 	active_power_usage = 500
 
 	var/uses = 0
-	var colour = "grey"
+	var/colour = "grey"
 	var/dirty = 0
 	var/obj/item/clothing/C = null
 	var/list/allowed_devices = list(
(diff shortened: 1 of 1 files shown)


LongSleeves:
Noticed Linter chan is getting pissy with how I declare and use my variables again. Kinda need uses, colour and dirty to be there. Will ask about it later on when I'm not dead tired.

code/game/machinery/xenobio.dm
```diff
@@ -70,18 +70,71 @@
 	var uses = 0
```
r4d6, line 70:
```suggestion
	var/uses = 0
```

code/game/machinery/xenobio.dm
```diff
@@ -70,18 +70,71 @@
...
 	var colour = "grey"
```
r4d6, line 71:
```suggestion
	var/colour = "grey"
```

code/game/machinery/xenobio.dm
```diff
@@ -70,18 +70,71 @@
...
 	var dirty = 0
```
r4d6, line 72:
```suggestion
	var/dirty = 0
```

PR #3165, opened 2022-04-20
Add Radiation Subsystem

r4d6:
Rework the current radiation """system""" to a ""proper"" radiation subsystem.
Also add Geigers Counters, which mesure the amount of radiation a certain tile it receiving, and by extension, how many rads any person standing on it receive. The amount can be seen by examining the Geiger Counter. Note : The Geiger Counter does not make any sounds when it start detecting radiation.


code/controllers/subsystems/radiation.dm
```diff
@@ -0,0 +1,79 @@
...
+		for(var/mob/living/M in range(L["Range"], L["Source"]))
+			M.apply_effect(L["Power"], IRRADIATE)
+			M.updatehealth()
+
+		// Run through all the geiger counters in range and update their radiation levels
+		for(var/obj/item/device/geiger/G in range(L["Range"], L["Source"]))
+			G.add_rads(L["Power"])
+
+// "Manually" irradiate nearby people. S is the source, P is the amount of rads, and R is the range
+proc/produce_radiation(var/S, var/P, var/R)
```
lolman360, line 56: probably could use better var names here

r4d6:
It has been tested on my side with the ameridian crystals as the tested source of radiation.

code/controllers/subsystems/radiation.dm
ThePainkiller: I concurr, using just letters can sometimes bring conflicts with other vars, just a neatness-of-code thing

PR #3169, opened 2022-04-21
silencer rebalance

lolman360:
## Changelog
:cl:
balance: silencer AP -10% >> 0%, silencer recoil -10% >> 0%
/:cl:


ThePainkiller:
It was always laughable for me that, videogame-wise, silencers always were a "debuff" for the gun, honestly love it being just "silent pew pew" as it should
Hell the aural dampener works the same, why not

lolman360:
aural dampener is actually worse since you need mod slots on tools- mod slots on guns aren't nearly as in high demand

if silencer becoms the be-all end-all muzzle slot option (i sincerely doubt it) i might add some more to the pool, or shift things like ported/gravcomp barrel to muzzle slot

PR #3179, opened 2022-04-23
Reductor Rebalance & Misc change to Bat'Ko + Adjudicator 

SouthyOsmanli:
## About The Pull Request
<details>
<summary>
	Long-awaited Reductor rebalance. Simply increased the charge cost on the Reductor (making the norm large capacity cell shoot 10 rounds instead of 20) and made it unable to take energy attachments - I would have blacklisted the attachments instead but the blacklist system refuses to work. Similarly the Abdicator is no longer able to take ballistic mods to stop gauss barrel from stacking with it, and the Bat'ko has received 1 attachment slot back since it can also now only take ballistic mods.
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
balance: Reductor is no longer able to take laser/energy modifications & has received a increase in charge cost (1000<500)
balance: Abdicator can no longer take ballistic modifications.
balance: Bat'ko has received back one of its attachment slots as it can no longer take energy attachments.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/abdicator.dm
+++ b/code/modules/projectiles/guns/energy/abdicator.dm
@@ -27,7 +27,7 @@ It also has more matterals then it takes to craft as a way to have a sunk cost.
 	charge_cost = 150
 	recoil_buildup = 12 //Big blast of heated shot, got some recoil to it.
 	damage_multiplier = 1 //already quite a bit lethal and dangerous with the burn damage and 'close range spray'.
-	gun_tags = list(GUN_PROJECTILE, GUN_LASER, GUN_ENERGY, GUN_SCOPE) //essentially a scattershot reductor.
+	gun_tags = list(GUN_LASER, GUN_ENERGY, GUN_SCOPE) //essentially a scattershot reductor.
 
 	proj_step_multiplier = 0.5 //2x bullet speed, mostly for flares so they dont crawl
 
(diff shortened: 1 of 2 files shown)


ThePainkiller:
Halving the rounds, I get, but jesus no energy adjustments? Can the reductor be actually properly balanced for once or will we be back to the "this weapon sucks again so no one is gonna build it" perpetual cycle?

SouthyOsmanli:
> Halving the rounds, I get, but jesus no energy adjustments? Can the reductor be actually properly balanced for once or will we be back to the "this weapon sucks again so no one is gonna build it" perpetual cycle?

The shunt alongside other attachments lets it do 150 damage with bonus AP as well. You are firing slugs, from a shotgun, that have double the affective range before their damage drops-off. The other option is hard nerfing energy mods or blacklisting certain ones.

Blacklist code does not work currently, nor does my work-around, and nerfing energy mods would just be declined as a PR due to other energy weapons needing it.

BigBadPossum:
The rail rifles unique and interesting feature is its ability to take multiple mod types or, essentially, any mod type. I feel as though it is balanced where it is. The reductor is a high end and incredibly powerful weapon, sure, but its limited to guild and dungeon rewards only and to truly shine as a weapon of class it needs mods, otherwise it is a shotgun with high capacity and worse munitions pooling.

PR #3201, opened 2022-04-27
bows and arrows, or How I Learnt To Stop Worrying And Love The Lodge

lolman360:
adds bows, which use the charge mechanic from eguns

charge mechanic can now be used for non-energy guns

support for supereffective damage on projectiles, allowing true 'hunting' weapons without overstacking AP

new mods for bows (just 2 atm, fast charge less damage and slow charge more damage

wip, still need sprites

## Changelog
:cl: lolman360, Churro
add: bows, bow mods
refactor: minor projectiles/ charge firemode tweaks
/:cl:


ThePainkiller:
Would be nice if bow/crossbow projectiles had a different sound than the throwing one though...

code/modules/mob/living/living_defense.dm
```diff
@@ -125,6 +125,8 @@
 		hit_impact(P.get_structure_damage(), hit_dir)
 		for(var/damage_type in P.damage_types)
 			var/damage = P.damage_types[damage_type]
+			if(faction in P.supereffective_factions)
+				damage *= P.supereffective_mult
```
hex37, line 129: Hey, is there any way you could get this supereffective mult outside of our bullet_act proc? An O(n) search loop here will probably slow things down a little.

lolman360: I can definitely do something like
```
if(LAZYLEN(P.supereffective_factions))
    loop
```

cdb-is-not-good:
Cool! Serious balance concerns with this, however. If I am reading right, these numbers are very high - and have super_effective bonuses against neutral, which...is every player,.

ThePainkiller:
Could we get the very base _"simple bow"_ parent on a crafting recipe available for anyone on the guns tab?

Would also be good if they could be worn on the suit slot, instead of having to take only the backpack slot, as you can still wrap bows around your chest.

lolman360:
> Cool! Serious balance concerns with this, however. If I am reading right, these numbers are very high - and have super_effective bonuses against neutral, which...is every player,.
balance is based around bonerifle which has similar crafting costs (20 bone 10 plasteel vs 10 bone + nightmare stalker special loot)- it's around as wieldly

assumed players are a separate faction, fixed in next update, it's now typepath based
supereffectivity re: factions is being moved to the `item` var `effective_factions` and `damage_mult` since projectiles are items

lolman360:
> Could we get the very base _"simple bow"_ parent on a crafting recipe available for anyone on the guns tab?

done

> Would also be good if they could be worn on the suit slot, instead of having to take only the backpack slot, as you can still wrap bows around your chest.

they can be worn in suit slot

cdb-is-not-good:
> balance is based around bonerifle which has similar crafting costs (20 bone 10 plasteel vs 10 bone + nightmare stalker special loot)- it's around as wieldly

I'll have to give it a test to see. The numbers feel fairly high and it's worth noting that if the bows are balanced around the baroque(a hunter exclusive) without being hunter exclusive, then it gives a once powerful tool of the hunters to many. But I'll playtest it tonight when I get home just to see

cdb-is-not-good:
Playtested a bit. Inhand for bows needs work. Should change the firing decal so that it's not the same as the guns(looks weird.).
Bows should be two handed, not one handed - for the sake of sanity. Holding bow in one hand, loading arrow, and drawing string back is...ehhhh? I stand by what i said about serrated arrows and their damage values, this is DEFINITELY not "as wieldy as the Baroque" - which requires open bolt, load shot, two hand, fire, unhand, open bolt - etc.

lolman360:
>The numbers feel fairly high and it's worth noting that if the bows are balanced around the baroque(a hunter exclusive) without being hunter exclusive, then it gives a once powerful tool of the hunters to many. 

hunting arrows and the recurve bow are lodge exclusive crafts


> Playtested a bit. Inhand for bows needs work. Should change the firing decal so that it's not the same as the guns(looks weird.)

easy enough

>Bows should be two handed, not one handed - for the sake of sanity. Holding bow in one hand, loading arrow, and drawing string back is...ehhhh? I stand by what i said about serrated arrows and their damage values, this is DEFINITELY not "as wieldy as the Baroque" - which requires open bolt, load shot, two hand, fire, unhand, open bolt - etc.

they should be twohanded, i might've changed it back for testing. i personally found drawing back to be a little more restrictive and absolutely more time-consuming than opening/closing bolt

you haven't actually said anything about serrated arrows yet, but 65 damage 0 ap might be a little much. all values are still up in the air at this point, but for reference serrated arrows are balanced ~around slugs where all the ap was traded for damage. they have a comparable range (5 vs 6)

there's the thing about ammo being renewable, which I probably haven't taken into account re:powerbudgeting

cdb-is-not-good:
Sounds good, we'll iron  out the more serious nitty-gritty of balance once it's actually in play and we can see how they get used. Once the conflicts are resolved lmk and we'll merge this - barring any further points of balance.

lolman360:
Won't be good to merge until I get a sprite for the Guild compound bow, but I'll resolve the conflicts soon. Sprites in general are a little rough IMO, but I can't do any better.

Dongels:
I'm like 90% sure Nayu had some bow sprites..or at some point we had someone post some good looking ones on discord...might of been Gen now that I think about it...hmm.   *shrug*

ThePainkiller:
Suggestion: Back slot icon states

lolman360:
waiting on sprites

ThePainkiller:
Why is the back icon a crossbow though?

image

lolman360:
> Why is the back icon a crossbow though?
> 
> image

used wrong file when resolving conflicts

PR #3211, opened 2022-04-28
BS Teleporter - +Blackguard, -Strelki, +Duty, Adds .408 Drum

SouthyOsmanli:
## About The Pull Request
<details>
<summary>
	Simple changes to the Blackshield teleporter prices. The Blackguard at only 300 power is incredibly cheap for one of the best guns valued at over 2500 cred. (1.35 dam mod, 1.2 pen mod - single, burst, etc) - increased it since armory starts with 2 and these should be harder to get. Decreased the Strelki price slightly since it kind of is a budget Watchman. Increased the Duty's price a tiny bit since it was equal to SMGs despite its decent damage. Added .408 drum mags since the Blackguard and the STS-40 takes them. Gives the STS-40, especially, a slight 'edge' over the Omni-platforms with ability to get a drum mag on it.
</summary>
<hr>
	
<hr>
</details>

## Changelog
:cl:
add: Adds .408 Drum to BS Teleporter.
balance: Blackguard power cost greatly increased, Strelki power cost slightly decreased, Duty power cost slightly increased.
/:cl:


ThePainkiller:
1000 power for a Blackguard is absolutely **ruthless**, the weapon is worth IMHO at least 600 power just to make it clear it's a soft "upgrade" to the STS-40 (pre-scoped, pre-bayo, literal 0.1 more values on multipliers on both damage and pen), not nearly 95% of the price of a Blackshield SCAF armor. Even with a price of 300, I've always personally opted for a STS over it due to 2 more mod slots, and being able to fit a better scope than just five more tiles.

Speaking of the Blackguard, its `update_icon()` proc is not working with drum mags, never has in fact (even before the drum capacity nerf, it only displays the overlay for normal mags but not for drums.

cdb-is-not-good:
While i do agree 1000 for the blackguard is fairly rough it's also a VERY good gun i m h o. It might need to be tweaked down but it's a weapon that's a very hard upgrade from an STS-40 given that it is a weapon they start with/can obtain through the teleporter. I'd be open to reducing the price at a later time if it turns out to be too good, but given what i talked to with Das about potentially making the 'omni-platforms' a little harder to get and overall staying at their power level i think it should be fine.

PR #3224, opened 2022-04-30
New Drip

Jul-Jul:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

This adds PR adds the skinsuit for public use including a grey and blue version for the two different body types which can be adjusted via style switching them. Grey turtleneck with jeans or skirt option adjusted via style verb. Marshal turtleneck now having a skirt version too.  

ALSO

All Sprites used were made by Kavrick and are not my own!
	
<hr>
</details>


cdb-is-not-good:
SHOULD work now, also SHOULD fix the toga too.

PR #3225, opened 2022-05-01
stack autostacking, haul on controlclicking tiles

lolman360:
Tin. Stacks autostack when crossing a stack that they can stack with.
Controlclicking a tile calls Haul on that tile.
## Changelog
:cl:
add: Stacks autostack when crossing a stack that they can stack with. Controlclicking a tile calls Haul on that tile.
/:cl:


ThePainkiller:
Wait a minute, give me hope please

Does this mean we finally have the tg stuff where you drag a stack of metal/floortiles on top of other sheets of the same material and they automatically stack??!?!

PR #3229, opened 2022-05-02
ROUND 2 - Railrifle & Abdicator Balancing

SouthyOsmanli:
## About The Pull Request
<details>
<summary>
	Simply put - the Railrifle that the Guild could make was incredibly broken as a weapon (ex builds allowing for roughly 150 damage and +30 AP) since it fires shotgun slugs at an affective range of 10 tiles before any damage/pen dropoff occurs. While this may have been fine if the projectiles were lasers, thus unable to break bones, the brute projectiles had effectively no counter as hit-scans. Due to this I've re-did my older PR - though I realized a charge increase on the railrifle was indeed too punishing. Now due to Lolman's previous PR, **BLACKLISTING WOOOORKS!!** Now it just can't take shunts, more attachments will be added as needed rather easily. Secondly, the Abdicator was a bit broken when equipped with a void wolf barrel, so I've disabled ballistic attachments for it since it fires energy-base projectiles anyway. Shouldn't effect it much though, it's still very powerful - only attachment I noticed made a big difference was the void wolf barrel. **Tl;dr** - Guild may hate me but, as a Guildie player, the rail is busted and really needs a 'soft nerf'.
</summary>
<hr>
	
<hr>
</details>

## Changelog
:cl:
balance: Railrifle and it's child-paths now has a working blacklist for energy-shunts. Also - lost its laser attachments.
balance: Abdicator lost its ability to take ballistic attachments (Why you'd put any besides a wolf-barrel on it anyway, I don't know.)
balance: Mild buff of the Bat'ko's scrap capacity since cells are small. (From 4 scrap to 6. Still has to vent every 2 rounds or else overheat.)
/:cl:


lolman360:
myrmidion taking kurtz kits makes no sense imo, since it's meant to bring other .50kurtz guns up to myrmidion tier

this makes myrm even more worthless while nerfing reductor into an actually balanced state for once

cdb-is-not-good:
Kinda weird choice on giving the Mymidon the kurtz kit but that's not a deal breaker imho.

SouthyOsmanli:
> myrmidion taking kurtz kits makes no sense imo, since it's meant to bring other .50kurtz guns up to myrmidion tier
> 
> this makes myrm even more worthless while nerfing reductor into an actually balanced state for once

It fires a .50 kurts round. Not allowing it to take Kurtz kits makes it effectively inferior to any .50 pistol. If no shunt, and no .50 kurts kit - thing might as well never be used.

lolman360:
blacklists now work so

SouthyOsmanli:
Blacklisting now added, in turn gave back energy attachments. Laser attachments don't fit still since that's lenses anyway and it - shouldn't be able to take lenses.

lolman360:
itt: let them still take unmakers, they're rare enough and these guns are hardly the best option to use them on

PR #3240, opened 2022-05-03
50% reduction in beacon entropy

lolman360:
wow this is problematic due to bulk exports


cdb-is-not-good:
90% might be...a bit too much. I'd suggest Eris guidelines of no more than 50% nerf or buff and lower it to like....one.

lolman360:
> 90% might be...a bit too much. I'd suggest Eris guidelines of no more than 50% nerf or buff and lower it to like....one.

done

PR #3255, opened 2022-05-05
SST balancing

cdb-is-not-good:
SST guns nerfed, but also not that badly.
SST weapon no longer work through all armor, they now instead work much akin to pepperball ammo(checks face covering, essentially. Must be something akin to a gas mask to be defended against). They retain their rubber rounds and still punch high above their weight.
Additionally they were given some slightly adjusted 'effective range', to give a distinct advantage over just using a /real/ gun using similar rubber ammo.
Abnegate : affective_damage_range = 2->5
Formatbound : affective_damage_range = 4->6
Humility: affective_damage_range = 1->5
Systemcost : unchanged.
Their price in the protolathe however has been reduced. They no longer require uranium, and at request instead of fully removing their platinum it has been SIGNIFICANTLY reduced(.1, .2, .5 for the abnegate, formatbound, humility/systemcost respectively). Their silver/gold costs have been left alone.

The idea of these changes is to make the SSTs a little more accessible and a little less of an 'i win' button with no real counter. They'll still OFTEN be a preferable option over cheaper, more readily available options, without being....so broken.

ALSO emergency nerfs the entropy for the beacon from 1 to .25, @KV-90 "Bear" ICBM#9021 was accurate in his initial guess about what was a valid number for the gain :V

### DIFF
--- a/code/modules/projectiles/guns/energy/sst.dm
+++ b/code/modules/projectiles/guns/energy/sst.dm
@@ -14,7 +14,7 @@
 	w_class = ITEM_SIZE_NORMAL
 	force = WEAPON_FORCE_NORMAL
 	origin_tech = list(TECH_COMBAT = 10)
-	matter = list(MATERIAL_PLASTEEL = 12, MATERIAL_STEEL = 8, MATERIAL_SILVER = 8)
+	matter = list(MATERIAL_PLASTEEL = 12, MATERIAL_STEEL = 8, MATERIAL_SILVER = 8, MATERIAL_PLATINUM = 0.1)
 	charge_cost = 100
 	twohanded = FALSE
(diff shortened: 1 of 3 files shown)


SouthyOsmanli:
Firstly there is absolutely no reason for this to be spawning, secondly it will screw up loot and make all these items more common, thirdly, this is possibly a direct violation to contributing guidelines, I should try reading them. https://wiki.cev-eris.com/Content_GuidelinesEn

r4d6:
> Firstly there is absolutely no reason for this to be spawning, secondly it will screw up loot and make all these items more common, thirdly, this is possibly a direct violation to contributing guidelines, I should try reading them. https://wiki.cev-eris.com/Content_GuidelinesEn

Uh... I don't want to be disrepectful, but since when do we use **_Eris's_** contribution guidelines ?

PR #3296, opened 2022-05-09
So long Stomper

cdb-is-not-good:
## About The Pull Request
<details>
<summary>
Replaces the venerable old stomper with the frankly equally venerable, but more useful M.A.I.D - Marshals Armored Impeller Device.
A stripped down gygax with worse energy efficiency per step, worse armor, less health, easier to break internals, less deflect chance and slightly more speed. Broadly speaking, it's not quite as good as bringing a real gygax to a fight but I hope that it will be capable of 'riot suppression' like Stomper was supposed to be.  Additionally, instead of coming with a giant MURDER sword it instead comes with a heavy-duty exosuit shock maul that while still dangerous will generally not kill unless you're TRYING to kill(Less lethal not NON lethal.)
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


cdb-is-not-good:
Not actually mapped in yet, wanted to get balance feedback first.

cdb-is-not-good:
> Not actually mapped in yet, wanted to get balance feedback first.

Mapped in, all good to go pending any balance concerns.

code/game/mecha/combat/gygax.dm
```diff
@@ -20,6 +20,41 @@
 	max_equip = 5
 	cargo_capacity = 3
 
+/obj/mecha/combat/gygax/marshals
+	name = "M.A.I.D Exosuit"
+	desc = "The Marshals Armored Impeller Device, a fancy name for a visibly aging exosuit that has visibly been collecting dust in some unused corner of the Marshals offices. \
+	It appears to have much lighter armor than even a standard gygax and despite still bearing the charictaristic angled plates of its Gygax cousin, is unlikely to stand up under extended\
```
ThePainkiller, line 26:
```suggestion
	It appears to have much lighter armor than even a standard gygax and despite still bearing the characteristic angled plates of its Gygax cousin, is unlikely to stand up under extended\
```

ThePainkiller (changes requested):
Smol typofix

ThePainkiller:
Also to have the acronym make sense, why not call it _"**M**arshals' **A**rmored **I**ncentive of **D**ispersal"_? Since it's supposed to be for riot control, I feel it fits better to call it an "incentive" to "disperse" rowdy crowds!

r4d6:
Currently, the MAID has the same armor as a regular gygax, which is determined by the `damage_absorption ` variable, which is a list.


Now, be careful, because [as precised in the variable's definition](https://github.com/sojourn-13/sojourn-station/blob/master/code/game/mecha/mecha.dm#L51) : 
> //the values in this list show how much damage will pass through, not how much will be absorbed.

So, with the gygax's values of `damage_absorption = list("brute"=0.75,"fire"=1,"bullet"=0.8,"energy"=0.85,"bomb"=1)`, to have the MAID have worse armor, you would have to have the brute higher than 0.75, the bullets higher than 0.8, and energy higher than 0.85

cdb-is-not-good:
Good point, those values were updated to bring its armor in line with the alleged 'aging' that it's undergone.

PR #3318, opened 2022-05-12
adds 60 units as an option to chem dispensers 

donut25-1:
adds the option to dispense 60 units to the chem dispenser


ThePainkiller:
God I wish I would've seen this PR sooner to also ask for 15 units

PR #3321, opened 2022-05-12
hollowpoint rework, post-penetration damage

lolman360:
adds post-pen damage to hollowpoints. nerfs hollowpoint agony by ~66% across the board.

## Changelog
:cl:
balance: hollowpoints deal greater post-penetration damage
/:cl:


ThePainkiller:
Why add a "post pen" value when hollow points are, by definition, non-pen ammunition meant to deal with unarmored targets...?

lolman360:
> Why add a "post pen" value when hollow points are, by definition, non-pen ammunition meant to deal with unarmored targets...?

hollowpoints, by design, are worse at penetrating armor but do more damage when they **do** penetrate. post-penetration damage is a reflection of this, and allows us to have weapons that do more damage against unarmored targets without doing something like a total armor rework with negative penetration

cdb-is-not-good (changes requested):

cdb-is-not-good (approved):

cdb-is-not-good:
Seems to be all working, well test on live and tweak from there the numbers.

PR #3322, opened 2022-05-12
Bar drink dispenser changes 

donut25-1:
Moved milk , grape soda, watermelon juice and berry juice to there respective dispensers round start instead of behind upgraded parts. Milk to the coffee dispenser and the rest to the soda dispenser.


ThePainkiller:
The reason Milk was gated behind upgrades was to force players to source it out of Betsy, fostering co-departamental dependancy on a botanist and making it a limited commodity rather than a one press of a button, that's why it's gated behind the **highest** tier of upgrades. Then again, the coffee dispenser has cream roundstart, so I don't think it is that much of an issue considering most coffee recipes involve milk in some way.
No issues about the juices being moved tbh

PR #3327, opened 2022-05-13
BIG combat-mech weapon/equipment changes

cdb-is-not-good:
In summary
fixes a bug where one could spamclick to get multiple bursts out from mech weapons all at once.
mech energy weapons costs to use significantly increased and fire rate increased
multi-las cannon delay between shots radically increased so both shots don't come out instantly.

All mech weapons(unless i missed some) given proper material costs so they aren't just 15 steel in the lathe.
Ditto on mech armor, now actually costs plasteel
Mech ballistics adjusted.
UltraAC fires its burst a bit faster and has less delay between bursts,
scattershot adjusted to no longer use .40 and instead uses 'heavy' shrapnel rounds with more pellets, slightly more damage and increased steps between losing pellets - reduced burst from 4 to 1
flak cannon now fires proper buckshot, and only one volley per burst.
TO DO
--------
Deviation is currently not working at all, which is a problem for mech balancing across the board, needs to be fixed.
Mechs by and large will likely need a closer look at their maneuverability and durability, though the latter concerns me less given their severe weakness to explosions/emps

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/game/mecha/equipment/weapons/ranged/bullet_weapons.dm
+++ b/code/game/mecha/equipment/weapons/ranged/bullet_weapons.dm
@@ -104,15 +104,16 @@
 /obj/item/mecha_parts/mecha_equipment/ranged_weapon/ballistic/scattershot
 	name = "\improper LBX AC 10 \"Scattershot\""
 	icon_state = "mecha_scatter"
-	equip_cooldown = 15
-	projectile = /obj/item/projectile/bullet/magnum_40
+	equip_cooldown = 25 // we fire fairly slow, but do a LOT of damage up close.
+	matter = list(MATERIAL_STEEL = 25, MATERIAL_PLASTEEL = 15)
+	projectile = /obj/item/projectile/bullet/pellet/shotgun/scattershot //snoflaek projectile. works similarly to regular shotgun rounds with 2 more pellets and 2 more damage per pellet
 	fire_sound = 'sound/weapons/guns/fire/shotgunp_fire.ogg'
 	fire_volume = 80
 	projectiles = 40
 	max_ammo = 40
-	projectiles_per_shot = 4
-	fire_cooldown = 2 //The scatter shot is meant to be fast but doing it all at once makes it a bit buggy and leaves flying bullets
-	deviation = 0.7
+	projectiles_per_shot = 1
+	fire_cooldown = 1
+	deviation = 0.7 //currently DOES NOT work, will adjust later if it is fixed.
 	ammo_type = "12g"
 
 /obj/item/mecha_parts/mecha_equipment/ranged_weapon/ballistic/scattershot/loaded
@@ -123,12 +124,12 @@
 	desc = "The design of this weapon brings a whole new meaning to the term scrap cannon."
 	icon_state = "mecha_makeshift_scatter"
 	equip_cooldown = 15
-	projectile = /obj/item/projectile/bullet/magnum_40/scrap
+	projectile = /obj/item/projectile/bullet/pellet/shotgun
 	fire_sound = 'sound/weapons/guns/fire/shotgunp_fire.ogg'
 	fire_volume = 80
 	projectiles = 30
 	max_ammo = 30
-	projectiles_per_shot = 6
+	projectiles_per_shot = 1
 	deviation = 0.9
 	required_type = list(/obj/mecha/combat, /obj/mecha/working, /obj/mecha/working)
 	price_tag = 700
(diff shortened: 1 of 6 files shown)


code/game/mecha/equipment/weapons/ranged/bullet_weapons.dm
```diff
@@ -234,6 +236,7 @@
 	name = "\improper SRM-8 missile rack"
 	icon_state = "mecha_missilerack"
 	projectile = /obj/item/missile
+	matter = list(MATERIAL_STEEL = 20, MATERIAL_PLASTEEL = 10, MATERIAL_SILVER  =5)
```
r4d6, line 239:
```suggestion
	matter = list(MATERIAL_STEEL = 20, MATERIAL_PLASTEEL = 10, MATERIAL_SILVER = 5)
```

code/game/mecha/equipment/weapons/ranged/laser_weapons.dm
```diff
@@ -50,34 +52,42 @@
...
 /obj/item/mecha_parts/mecha_equipment/ranged_weapon/energy/ion
-	equip_cooldown = 40
+	equip_cooldown = 40 //this is fine. This is a very dangerous tool.
 	name = "mkIV ion heavy cannon"
 	icon_state = "mecha_ion"
-	energy_drain = 120
-	projectile = /obj/item/projectile/ion
-	fire_sound = 'sound/weapons/Laser.ogg'
-	price_tag = 300
+	/obj/item/mecha_parts/mecha_equipment/ranged_weapon/energy/laser/heavy
```
r4d6, line 67:
```suggestion
/obj/item/mecha_parts/mecha_equipment/ranged_weapon/energy/laser/heavy
```

r4d6: The weapon was already defined above the multi-laser

code/game/mecha/equipment/weapons/ranged/laser_weapons.dm
```diff
@@ -50,34 +52,42 @@
...
+	equip_cooldown = 15
+	name = "\improper CH-LC \"Solaris\" laser cannon"
+	icon_state = "mecha_laser"
+	matter = list(MATERIAL_STEEL = 25, MATERIAL_PLASTEEL = 15, MATERIAL_SILVER  =10)
+	energy_drain = 500 //same as multi-laser, VERY dangerous and VERY powerful
```
r4d6, line 72: The Multi-laser cost more power to operate because it is literally 4 of these stuck together to compensate for their individual low firerate.

code/game/mecha/equipment/weapons/ranged/laser_weapons.dm
cdb-is-not-good: oh huh, must've accidentally moved something about

code/game/mecha/equipment/weapons/ranged/laser_weapons.dm
cdb-is-not-good: Ah...this was also a mistake, that comment was meant to be over the ion cannons energy change. Will need to look over this in the morning and see what else i goofed up on

code/game/mecha/equipment/weapons/ranged/laser_weapons.dm
```diff
@@ -50,34 +52,37 @@
 	name = "\improper Multi-laser Cannon"
 	desc = "Four solaris laser cannons welded together to achieve a higher firerate."
 	icon_state = "mecha_auto_laser"
-	energy_drain = 180
+	matter = list(MATERIAL_STEEL = 10, MATERIAL_PLASTEEL = 5, MATERIAL_SILVER  =1)
```
r4d6, line 55:
```suggestion
	matter = list(MATERIAL_STEEL = 10, MATERIAL_PLASTEEL = 5, MATERIAL_SILVER = 1)
```

lolman360:
why are energy weapons getting across-the-board 3-5x cost increases, are they unbalanced in their current state or

cdb-is-not-good:
> why are energy weapons getting across-the-board 3-5x cost increases, are they unbalanced in their current state or

Yes, because exosuits start with a 16k cell for free. the costs were untouched after we went ahead and implemented 'ammo' for ballistics and while i don't want to go so far as to make them take like...a second battery or some silly shit, i certainly do want their fire-values to have SOME kinda impact on their use.

DiscordWizard:
Try to not overdo the energy weapon firing costs, even with their default cell charge every shot of a laser cannon takes 0.35% of the cell charge away. May not sound like much but that laser cannon will not kill anything without shooting it like 5 or 10 times and never missing a shot. 

That's 2-3% charge per mob on the old power drain. Considering mechs are used to fight either bugs or in events, killing 5 non maint bug mobs or 10 of them is already a 10% cell charge gone. Using energy weapons sacrifices alot of staying power over not carrying ammunition, and is honestly only viable if you have a self recharging cell.

I agree with wanting to make them a little more expensive but I _really_ don't want to spend every single event or fight having to take long pauses to let my cell recharge or swapping it every few minutes of using an energy weapon

cdb-is-not-good:
Mmm- My only concern is that raising it just an inch or two and then eventually maybe having to raise it again later is the kinda thing that rubs people wrong. But I do see yer point. For now i'll put them down by a bit.

PR #3360, opened 2022-05-17
Dylo OD + Mech Refining

DiscordWizard:
## About The Pull Request
<details>
<summary>
	Gives dylovene a harsher OD than just some toxins
	Makes it so mech ranged weapons properly melee things
        Cleans mecha code a little
</summary>
<hr>

Fixes me being dumb with the ranged weapon melee check on my earlier commit, I forgot to test hitting another mech in melee with a ranged weapon and it's most likely borked.

Changes Dylovene OD from mild toxins to a proper amount of toxic buildup and kidney damage, because organs that aren't the liver should receive some love too! And some gasping, it's always a good indicator
	
<hr>
</details>

## Changelog
:cl:
add: dylo OD limit at 30 units
balance: harsher dylo OD
fix: mech ranged melee fix
code: removed duplicated code from mecha
/:cl:


DiscordWizard:
Forgot to test this at home, so putting it up for review to see if anyone spots any errors or wants to comment and argue about the new dylo OD for balance before I work back on it in 12 hours

ThePainkiller:
Dylovene never had, **nor should it ever have**, any kind of overdose. What's the reason behind this?

DiscordWizard:
It was on the to-do list that dylovene needed a different OD than the basic one. Every single chemical has a base OD that gives mild toxins when surpassed, the problem with dylovene is that it simply out-heals any damage its OD can give so people can go around with as much dylo in them as they please. It's already pretty hard to OD on dylo with the 30 unit cap and extremely good toxin removal, if you go over that it's honestly on you.

cdb-is-not-good:
> Dylovene never had, **nor should it ever have**, any kind of overdose. What's the reason behind this?

why would it not have an overdose?

DiscordWizard:
The parent /reagent overdose is set to 0 and /reagent/medicine doesn't change it, you have to specify the overdose level on the specific chem if I understand the code correctly

Dongels:
I've been *for* giving Dylo an OD that was not just the current base tox...since it's a nothing thing that gets cured by the dylo. This would just help diversify the use of other tox chems.

cdb-is-not-good:
> I don't think it's necessary to include it since it should inherit from parent, as every other reagent that doesn't have it specified :B

*Insert stephan-molyneux meme about arguments here*
That's not really a proper response and you know it.

ThePainkiller:
> Insert stephan-molyneux meme about arguments here
That's not really a proper response and you know it.

...W-what kind of response? I just thought it redundant to add something that I assumed the code handled for itself by inheriting from parent, I didn't even say it in a rude way, I think...?

DiscordWizard:
You're fine, it's a valid concern. I just feel like we do need a little more damage variety and dylovene can be very easily abused since you just have an infinite amount of it in sleepers.
And hey, sleepers don't even let you add Dylovene to 30u by default! You're physically unable to OD someone unless you purposefully jab three syringes in them

PR #3426, opened 2022-05-25
Another attempt at fixing perk harddels

nikothedude:
## About The Pull Request
Title, tries to clear a few more references I found

Logically, this shouldn't break anything. Realistically, it might.

## Changelog
:cl:
fix: Tries to fix a statclick harddel.
/:cl:


nikothedude:
This might also fix the thousands of runtimes I created in perk.dm

nikothedude:
> ;v please test your PRs moving forward, even if it's just to boot up a test server and join the round. This is something that even a little testing would've caught.

I did test this one actually! I just didn't play with a perk that would've removed itself.

PR #3432, opened 2022-05-27
spiders now apply less venom to things that can't be injected

lolman360:
tin

psi monsters/xenomorphs/greyson hypo drones are unaffected because 1. they should be stronger and 2. they'd feasibly be strong enough to pierce suits (xeno/psi)  or have a injector that ignores suits (greyson)

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/giant_spider/attack.dm
+++ b/code/modules/mob/living/carbon/superior_animal/giant_spider/attack.dm
@@ -6,5 +6,5 @@
 	if(isliving(A))
 		var/mob/living/L = A
 		if(istype(L) && L.reagents)
-			var/poison_injected = poison_per_bite * ((-100 * L.getarmor(targeted_organ, ARMOR_MELEE) + 100) - 40)
+			var/poison_injected = poison_per_bite * (-100 * L.getarmor(targeted_organ, ARMOR_MELEE) + 100)
 			L.reagents.add_reagent(poison_type, poison_per_bite)
--- a/code/modules/mob/living/carbon/superior_animal/giant_spider/attack.dm
+++ b/code/modules/mob/living/carbon/superior_animal/giant_spider/attack.dm
@@ -6,5 +6,5 @@
 	if(isliving(A))
 		var/mob/living/L = A
 		if(istype(L) && L.reagents)
-			var/poison_injected = L.can_inject(src) ? poison_per_bite * 0.3 : poison_per_bite
+			var/poison_injected = L.can_inject(src) ? poison_per_bite * 0.6 : poison_per_bite
 			L.reagents.add_reagent(poison_type, poison_per_bite)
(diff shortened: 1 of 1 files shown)


cdb-is-not-good:
Im...not a fan of this, for a wide variety of reasons. None more serious, than the balance implications of neutering spiders against 99% of targets.

lolman360:
spiders are being buffed significantly if mob ticklag fixes go through, this is mostly in response to that

ThePainkiller:
Even without the _"ticklag fix"_, this sounds like a sensible change that makes sense tbh

Wearing a SCAF would protect against poison (but not the bites themselves), since atm spiders can inject reagents where normal syringes cannot.

cdb-is-not-good:
> Even without the _"ticklag fix"_, this sounds like a sensible change that makes sense tbh
> 
> Wearing a SCAF would protect against poison (but not the bites themselves), since atm spiders can inject reagents where normal syringes cannot.

Yes...and as has been pointed out, a hypodermic needle isn't designed to inject through thick materials - JUST skin. Needles are relatively small and flimsy - compared to a viper, they are roughly the same size. A giant spider however? Is the size of a LARGE DOG. Their fangs would be MAGNITUDES bigger, their bite strength magnitudes stronger. 

But realism isn't really relevant here, what's relevant is balance - and this is not, in my view...balanced.

lolman360:
spider venom should be counterable in a way that isn't 'stay out of melee range' imo, and thickmaterial solves the problem since it's not on most armors- and generally, armors which do have it have slowdown

cdb-is-not-good:
> spider venom should be counterable in a way that isn't 'stay out of melee range' imo, and thickmaterial solves the problem since it's not on most armors- and generally, armors which do have it have slowdown

Shields. Dylovene.

lolman360:
dylovene doesn't prevent it but yeah shields are fair

cdb-is-not-good:
i mean, dylovene purges it from your blood

cdb-is-not-good:
If you can rework this to not be quite such a nerf to spiders, I'll take a look at it. Else this will likely be closed soon

lolman360:
@cdb-is-not-good done, they now block 70% of the venom so it's not preventable outside of shields

cdb-is-not-good:
70% for as little as wearing armor is still a massive nerf to spiders across the board at a time when mobs don't really 'need' such a nerf

lolman360:
> 70% for as little as wearing armor is still a massive nerf to spiders across the board at a time when mobs don't really 'need' such a nerf

it's now 40%

cdb-is-not-good:
As discussed last night, this applies to pretty much all armor

lolman360:
done

MSieghardt:
CLAMP the second half of your equation to have a minimum and you should be fine. 

The venom is designed intentionally to make it so that you can't just facetank them like a dork.

PR #3433, opened 2022-05-27
Increases mob reaction time by approximately 300%.

nikothedude:
## About The Pull Request
Makes it so that when a mob finds a new target it aggros instantly and tries to attack it.

BE WARNED. THIS HAS BALANCE IMPLICATIONS. This increases aggro rate by 3x, I think, which will make the strategy of ganking a lot harder. Also, makes trap spiders attack before you can get up in some circumstances. 

I started this as a bugfux but realized it is just a buff to most mobs. 

NOTE. I tested this. Mobs seem to aggro very well now, targetting works, no runtimes, but I can't profile it. For some reason I don't have access to the native profiller. I don't know why.

I personally support this PR, because I think ganking is a shit strategy, and if we are basing our balance off of it, that kinda sucks. If this causes issues, I feel the best course of action would be to instead change other things, not this PR. Although, I will revert it if it causes too many problems.

## Changelog
:cl:
balance: Mobs should now react 3x as fast.
/:cl:

### DIFF
--- a/code/game/objects/effects/mines.dm
+++ b/code/game/objects/effects/mines.dm
@@ -119,7 +119,10 @@
 	M.Weaken(3)
 	var/mob/living/carbon/superior_animal/temp_ref
 	temp_ref = new /mob/living/carbon/superior_animal/giant_spider/tarantula/emperor(src.loc)
+	temp_ref.Paralyse(20)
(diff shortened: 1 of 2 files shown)


cdb-is-not-good:
Good stuff.

PR #3445, opened 2022-05-28
Fixes mobs firing even when LOS is broken, updates our can_see proc to TG, and a few more misc changes.

nikothedude:
## About The Pull Request
Title.

Sadly, this might come with a slight performance cost, but I really doubt it'll be too bad.\

The new argument for LoseTarget() is inconsequential and can potentially be used by others in the future.

## Changelog
:cl:
fix: Mobs should now properly lose targetting after firing at a target outside of LOS.
/:cl:


nikothedude:
and yes tested, 0 runtimes and stuff

PR #3452, opened 2022-05-29
Ranged mobs now draw a semi-transparent yellow beam to their attack targets a bit before they fire. 

nikothedude:
## About The Pull Request
Title.

Tested, works, no runtimes, although might cause slight performance issues. Profiler reported the beam was capable of causing lag, at least roundstart, where everything is targetting eachother. I can probably make this only be drawn towards sapient things, so if you're worried, tell me and I'll change it.

The idea behind this PR is that since the telegraph text doesn't really reflect EXACTLY who is attacking, the only way to show it is to visibly indicate what is about to be attacked.

## Changelog
:cl:
add: Ranged mobs will now cast a yellow beam very briefly toward whoever they intend to attack.
/:cl:


nikothedude:
Btw I'm fine if this PR is rejected, I myself have my doubts of it. LIke, I don't think it's IMMERSION breaking, but it's a little iffy.

nikothedude:
atomizing

nikothedude:
https://user-images.githubusercontent.com/59709059/170895497-7c98c423-402f-4afc-a085-ffe96fa332f0.mp4

ThatGuyFromThatPlace:
While I don't particularly like this idea for all mobs- I feel like this code could very much go to use with a sniper enemy that can fire at you from really long distances off-screen with a laser sight tracking whoever they're aiming at (An idea that I've wanted to see for a really long time)

nikothedude:
> While I don't particularly like this idea for all mobs- I feel like this code could very much go to use with a sniper enemy that can fire at you from really long distances off-screen with a laser sight tracking whoever they're aiming at (An idea that I've wanted to see for a really long time)

Would be pretty decent for that yeah. Not going to make any changes to it, not until a maintainer comes along and leaves their feedback, though.

nikothedude:
AAAAAAAAAAAAAAAAAAAAAA

nikothedude:
reaping what i sow

cdb-is-not-good:
Well try this for now and if it sucks we'll remove it

PR #3453, opened 2022-05-29
Fixes some more mob bugs.

nikothedude:
## About The Pull Request
Title. 

Mobs will now wait for X amount of ticks if their target is out of sight. After that, they lose aggro.
Mobs will no longer lose aggro if they lose sight of you.
Mobs won't attack if youre out of their sight.

Rush retarget now works as intended.
Misc code improvements.

## Changelog
:cl:
tweak:Mobs will now wait for X amount of ticks if their target is out of sight. After that, they lose aggro.
fix:Mobs will no longer lose aggro if they lose sight of you.
balance:Mobs won't attack if you're out of their sight.
fix:Rush retarget now works as intended.
/:cl:


nikothedude:
btw i forgot to check runtimes on my last test of this uuugh

PR #3484, opened 2022-06-03
AAAAAAAAAAAAA

nikothedude:
## About The Pull Request
i love it when i miss a obvious problem with a pr

fixes dead emotes and following because my other fix didnt fix it

## Changelog
:cl:
fix: fixes dead emotes and following
/:cl:


nikothedude:
only putting it up so it can be merged if you want

nikothedude:
okay, i think it works, but it does revert my following fix i made

PR #3487, opened 2022-06-04
Add a ranged ameridian golem

r4d6:
## About The Pull Request
\- Add a new, ranged ameridian golem that shoot ameridian shards that can embed.
\- Make ameridian crystals give hallucinations to nearby psions when spawning golems.
\- Add a new gun that uses Solid Ameridian as ammo and fire the same type of projectile as the ranged golems.

It is a draft while waiting for sprites & a better name for the golem & the gun.
The gun & the golem were tested and confirmed to be working.


code/modules/projectiles/projectile.dm
```diff
@@ -43,6 +43,7 @@
 	var/taser_effect = FALSE //If set then the projectile will apply it's agony damage using stun_effect_act() to mobs it hits, and other damage will be ignored
 	var/check_armour = ARMOR_BULLET //Defines what armor to use when it hits things. Full list could be found at defines\damage_organs.dm
 	var/projectile_type = /obj/item/projectile
+	var/shrapnel_type = /obj/item/material/shard/shrapnel
```
lolman360, line 46: shrapnel type is intentionally null- if it's null, it uses shard behavior on embed, if it's non-null, it has special behaviour

r4d6: What are you on about ?!

This is so that we have other kind of shrapnel than steel shrapnel. And **everything** shrapnel-related is handled on the carbon mob. It was the only way for me to allow the projectile to embed as shrapnel without it magically turning into steel shrapnel.
Beside, I looked at the code, the shrapnel was previously _**hardcoded**_ to be steel shrapnel, and no exceptions were allowed.

r4d6: He answered on discord, he forgot his shrapnel changes weren't in yet.

r4d6:
They will not.

r4d6:
> > They will not.
> 
> You might want to increase base-damage slightly more as a result then, embeds themselves aren't particularly powerful. Perhaps a slight tad more AP? I dunno. Ameridium is pretty unreliable and tied to server performance, unlike ammo/etc which can be massproduced with t4 parts with just materials, and only a slight delay with lag.

It deal 40 damage with 25 AP, and while ameridian production is unreliable, it`s still easy to get hundreds of ameridian shards to use as ammo.

cdb-is-not-good:
Is this still being worked on?

r4d6:
> Is this still being worked on?

Just waiting for sprites for the golem & gun, which will probably never come though...
So it's basically ready to be merged and won't break anything if it is merged.

cdb-is-not-good:
mkay ,we can get sprites later i suppose

PR #3491, opened 2022-06-05
Remove department base budget, but keep the wages

r4d6:
\- Remove the base 500$ that each department would receive every hour, but still allow them to receive money for the wages from their respective factions/sponsor.
\- Also cut the starting money to 10% of what it previously was, so that they don't have enough money for every vanity project that pass through their mind without making money another way.
\- It fixes the wages not properly accounting for the Nepotism & Debtor perks.

[TRILBY, only put the above part in the changelog ^]
Suggestions for the amount of starting money/hourly budget is welcome, but I won't budge on the wage thing.

Alternative to https://github.com/sojourn-13/sojourn-station/pull/3446

### DIFF
--- a/code/game/jobs/department.dm
+++ b/code/game/jobs/department.dm
@@ -16,7 +16,7 @@
 	var/funding_source
 
 	// Budget for misc department expenses, paid regardless of it being manned or not
-	var/budget_base = 500
+	var/budget_base = 0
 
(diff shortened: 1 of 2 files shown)


Mocha-Not-Latte:
Honestly - I think down to 10% a little too much, maybe like 25-30%? Would be a bit better for the department accounts, might allow some things still but not all too much

PR #3493, opened 2022-06-06
Adds a var to shield generators that controls if it will report it's damage or not.

nikothedude:
## About The Pull Request
I want to use one of these in an event at some point abut I can't if they'll constantly scream about their damage status.

## Changelog
:cl:
tweak: Added a var to shield gens that allows them to not report it's damage.
/:cl:

### DIFF
--- a/code/modules/shield_generators/shield_generator.dm
+++ b/code/modules/shield_generators/shield_generator.dm
@@ -651,14 +651,15 @@
 	if (report_scheduled)
 		return
 
-	report_scheduled = TRUE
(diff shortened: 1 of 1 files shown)


cdb-is-not-good:
Is this variable controllable from the shield interface?

nikothedude:
> Is this variable controllable from the shield interface?

No actually, good idea, I'll do that when I get a chance

nikothedude:
g o d n o .

cdb-is-not-good:
>g o d n o .

lmao

PR #3495, opened 2022-06-06
More Mech Weapons: The Balance Edition

Jul-Jul:
## About The Pull Request
<summery>

This PR firstly adjusts the Autocannon and its jury rigged version taking into account of player given feedback, however would require someone to review the balance. Also removed the welding ability of the mech energy cutlass as it was preventing people from putting it on the mech...and pretty sure could be used to repair other mechs with your own mech due to the welding ability. Considering the limb plating that allows mechs to punch shit is considered a combat weapon and thus illegal without permit on a mech, I have increased its damage and material cost. 

<hr>
</summery>


cdb-is-not-good:
seems fine.

PR #3545, opened 2022-06-15
If a mob is dead, it will now clarify that

nikothedude:
## About The Pull Request
Title. Just adds another line of text that shows up if a mob is dead. Untested.

## Changelog
:cl:
tweak: Made it so mobs now clarify it they are dead or not
/:cl:


nikothedude:
FOR GODS SAKE

PR #3547, opened 2022-06-15
Makes mobs wander a little if they can't see you.

nikothedude:
## About The Pull Request
Title.

This will do two things: 1. If a melee mob is stuck against a wall, it might rub itself over to the left or right and find it's way through a doorway to it's target on the other side. 2. If a ranged mob's can_see is returning false, erronously, which does happen, it might allow it to reposition to a position that lets it actually hit it's target.

One is questionable, but two is just a bugfix.

	
<hr>
</details>

## Changelog
:cl:
balance: Mobs will now wander if they can't see you.
/:cl:


nikothedude:
oh yeah shit i forgot about the other changes

PR #3558, opened 2022-06-16
Mobs lose sight and stop trying to follow you if you break LOS. They instead follow your position. Also a few bugfixes and code improvements.

nikothedude:
## About The Pull Request
Title.

It was always weird to me how they followed you through walls. They don't do that anymore. They'll still chase down your most recent position, and will probably find you again, but now you can trick them by looping around.

Also fixes the bug with burst fire mobs firing when dead, improves some code, etc etc etc. I put the weakened check on handle_ai as that's the parent proc for many of the procs it was on.

## Changelog
:cl:
fix: Fixed burst fire mobs firing when dead
balance: Made mobs lose sight of you and chase your msot recent position if they lose LOS
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/ai.dm
+++ b/code/modules/mob/living/carbon/superior_animal/ai.dm
@@ -77,6 +77,7 @@
 	melee_delay = melee_delay_initial
 	target_mob = null
 	stance = HOSTILE_STANCE_IDLE
+	lost_sight = FALSE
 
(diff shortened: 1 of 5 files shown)


nikothedude:
and yeah tested and working woo

nikothedude:
though to be fair i didnt test as much as i usually do on this one, but mobs stopped following me through walls, retargetted if they saw me, lost target normally, etc

PR #3561, opened 2022-06-17
Makes pepper spiders spray pepper at enemies, and shock rock spiders shock carbons.

nikothedude:
## About The Pull Request
Title.

Less of a nerf or a buff, more of making them more unique. 

The reasoning for this is that pepper spiders lose some damage due to this, and are hard countered by gas masks-before, they did extra damage sometimes, that ignored armor, same for shock rock spiders. Shock rock spiders though, are honestly just, kind of buffed because of this. Electrocution is way better than oxyloss, both in the fact it actually does damage, and the fact it can sometimes stun.

I do think it's worth it, but I'm not quite so sure about the balance implications. I'm sure it'll be fine, though.

	
<hr>
</details>

## Changelog
:cl:
add: Pepper spiders will now spray pepper spray at their targets, and shock rock spiders will now electrocute carbons.
/:cl:


cdb-is-not-good:
didja test it in play to see how it actually interacts with the typical 'gear' one might wear? That'd be my suggestion for testing the balance implications

nikothedude:
> didja test it in play to see how it actually interacts with the typical 'gear' one might wear? That'd be my suggestion for testing the balance implications

not really, no, but

pepper spiders, logically, follow the same logic as pepper spray, so gas masks and glasses SHOULD block them. i tested it as a BST, my glasses blocked the pepper spray

however, electrocute_act() does not really care about if you're wearing anything insulated or not, so that damage will always get applied

not like it ever /didnt/, because shockrock spiders ignored armor on their special attack, but worth considering

nikothedude:
if youre not sure about this, i can run the pr through some more practical tests tomorrow

nikothedude:
> Main issue would be pepperspray spiders burrowing out and instantly killing someone because they weren't wearing a mask and thus get stunlocked by capsaicin until help arrives, which is the real issue with this.
> 
> Otherwise the shock part is good, maybe give it a cooldown on the spray?

Again, stunlocks are nearly impossible. In order for someone to be stunlocked, they'd have to be pepper sprayed on the exact tick they recover from their weakness and stun. It's a /possibility/, just rare. A cooldown wouldn't be amiss though, although the random chance already does that to an extent.

I'll do some local testing and see what I come up with.

nikothedude:
> > > Main issue would be pepperspray spiders burrowing out and instantly killing someone because they weren't wearing a mask and thus get stunlocked by capsaicin until help arrives, which is the real issue with this.
> > > Otherwise the shock part is good, maybe give it a cooldown on the spray?
> > 
> > 
> > Again, stunlocks are nearly impossible. In order for someone to be stunlocked, they'd have to be pepper sprayed on the exact tick they recover from their weakness and stun. It's a /possibility/, just rare. A cooldown wouldn't be amiss though, although the random chance already does that to an extent.
> > I'll do some local testing and see what I come up with.
> 
> Considering Capsaicin does extreme pain damage, even past the forced stun and knockdown duration you may still suffer stunlocking if you have low toughness. The cooldown shouldn't be too long, ~10 seconds per spray, maybe 15.

Ahhhh, fair, I didnt consider the pain

nikothedude:
ALright, I think this is good to merge, to an extent, I ran around as a colonist and let a pepper spider spray me, and I wasn't completely fucked, only moderately.

PR #3563, opened 2022-06-17
Revert "Health through Stats"

Vizoplez:
Reverts sojourn-13/sojourn-station#3528

This... completely changed our balance, specifically involving more aggressive storytellers (which now work again).

Look, I'm not going to sugarcoat it

What the actual fuck? I leave the dev-team for more than a week and Trilby massively changes how the game works again? Including decreasing the health of every non-combat role by 10? And then making it a stat?

I'm sorry - but what was the actual intent of this? What meth are you smoking?

### DIFF
--- a/code/__DEFINES/mob_stats.dm
+++ b/code/__DEFINES/mob_stats.dm
@@ -6,11 +6,8 @@
 #define STAT_ROB			"Robustness"
 #define STAT_TGH			"Toughness"
 #define STAT_VIG			"Vigilance"
-#define STAT_VIV			"Vivification"
-#define STAT_ANA			"Anatomy"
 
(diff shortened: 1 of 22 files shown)


nikothedude:
hi helloi did nto read anything in the pr but

the health changes always felt really really weird to me, why nerf people for playing a specific job?

nikothedude:
> Partially my point, but the bigger issue is that this wasn't atomised at all, so we either had to accept the PR as a whole, or deny it as a whole.

(((or deny the pr and demand atomization like we should be doing)))

nikothedude:
> > > Partially my point, but the bigger issue is that this wasn't atomised at all, so we either had to accept the PR as a whole, or deny it as a whole.
> > 
> > 
> > (((or deny the pr and demand atomization like we should be doing)))
> 
> why would I atomise their content for them?

DEMAND atomization

PR #3583, opened 2022-06-20
LightPink Slime Reactions

LongSleeves:
Light pink slime cores now produce pink slaughto-matics.
They also produce wood baseball bats.
Peace thru a iron fist~


cdb-is-not-good:
neat

PR #3618, opened 2022-06-29
a godawful dirty fix for mobs following you while dead

nikothedude:
## About The Pull Request
title. this fucking sucks. im at the limit of my sanity because i just discovered this entire issue is because of the fact that mobs are INCONSISTANTLY getting ai_inactive set to false from superior mob death to base death

its so goddamn cursed

no energy to test it, but it doesnt runtime at least

## Changelog
:cl:
fix: hopefully fixes mobs sliding while dead
/:cl:


nikothedude:
less godawful now

PR #3650, opened 2022-07-08
Laser ammo rebalance

Dimasw99:
Laser ammo has been adjusted to use the AP value of the standard ammunition type while using the damage value of the scrap ammo.

In its current state some laser ammunition variants have the issue where even the scrap bullet variant were performing better than the vastly more expensive laser variant. Even when there is no guild around there is little incentive to use this type of ammo over by simply handcrafting it using 9 steel and 1 cardboard. 

The main goal of this PR is to position laser ammo to be right between scrap and guild ammo. Giving them more of a reason to be used all while still leaving guilds ammo throne untouched. 

## Changelog
:cl:
balance: Laser ammo now use the AP of standard bullets all while using the damage of scrap bullets
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -46,25 +46,31 @@ In pvp they also have more lasting damages, such as infections, and pain form bu
 
 /obj/item/projectile/beam/weak/pistol_35
 	damage_types = list(BURN = 12)
+	armor_penetration = 5
 	recoil = 1
 
 /obj/item/projectile/beam/weak/light_rifle_257
-	damage_types = list(BURN = 13)
+	damage_types = list(BURN = 18)
+	armor_penetration = 15
 
 /obj/item/projectile/beam/weak/rifle_75
 	damage_types = list(BURN = 22)
+	armor_penetration = 20
 	recoil = 4
 
 /obj/item/projectile/beam/weak/heavy_rifle_408
-	damage_types = list(BURN = 16)
+	damage_types = list(BURN = 20)
+	armor_penetration = 30
 	recoil = 3
 
 /obj/item/projectile/beam/weak/magnum_40
 	damage_types = list(BURN = 20)
+	armor_penetration = 10
 	recoil = 3
 
 /obj/item/projectile/beam/weak/kurtz_50
-	damage_types = list(BURN = 40)
+	damage_types = list(BURN = 30)
+	armor_penetration = 15
 	recoil = 5
 
 /obj/item/projectile/beam/weak/smg
@@ -79,8 +85,8 @@ In pvp they also have more lasting damages, such as infections, and pain form bu
 	armor_penetration = 25
 
 /obj/item/projectile/beam/shotgun
-	damage_types = list(BURN = 35) //Normal slugs deal 45
-	armor_penetration = 10
+	damage_types = list(BURN = 48) //Normal slugs deal 45 - Not anymore! Even scrap dealt more with 48!
+	armor_penetration = 25
 	recoil = 2
 
 /obj/item/projectile/beam/shotgun/strong


code/modules/projectiles/projectile/beams.dm
```diff
@@ -46,25 +46,31 @@ In pvp they also have more lasting damages, such as infections, and pain form bu
 
 /obj/item/projectile/beam/weak/pistol_35
 	damage_types = list(BURN = 12)
+	armor_penetration = 5
```
Trilbyspaceclone, line 49: Why are you nerfing the pistol one?

code/modules/projectiles/projectile/beams.dm
```diff
@@ -46,25 +46,31 @@ In pvp they also have more lasting damages, such as infections, and pain form bu
...
 	recoil = 1
 
 /obj/item/projectile/beam/weak/light_rifle_257
-	damage_types = list(BURN = 13)
+	damage_types = list(BURN = 18)
+	armor_penetration = 15
```
Trilbyspaceclone, line 54: Same AP as the parrent but why more damage?

code/modules/projectiles/projectile/beams.dm
```diff
@@ -46,25 +46,31 @@ In pvp they also have more lasting damages, such as infections, and pain form bu
...
 	recoil = 3
 
 /obj/item/projectile/beam/weak/magnum_40
 	damage_types = list(BURN = 20)
+	armor_penetration = 10
 	recoil = 3
 
 /obj/item/projectile/beam/weak/kurtz_50
-	damage_types = list(BURN = 40)
+	damage_types = list(BURN = 30)
```
Trilbyspaceclone, line 72: Huge nerf thats really odd as few guns use this

code/modules/projectiles/projectile/beams.dm
```diff
@@ -46,25 +46,31 @@ In pvp they also have more lasting damages, such as infections, and pain form bu
...
+	armor_penetration = 15
 
 /obj/item/projectile/beam/weak/rifle_75
 	damage_types = list(BURN = 22)
+	armor_penetration = 20
 	recoil = 4
 
 /obj/item/projectile/beam/weak/heavy_rifle_408
-	damage_types = list(BURN = 16)
+	damage_types = list(BURN = 20)
```
Trilbyspaceclone, line 62: This will be way to op in pve as it deals 80 damage to mobs

code/modules/projectiles/projectile/beams.dm
```diff
@@ -46,25 +46,31 @@ In pvp they also have more lasting damages, such as infections, and pain form bu
...
+	armor_penetration = 30
 	recoil = 3
 
 /obj/item/projectile/beam/weak/magnum_40
 	damage_types = list(BURN = 20)
+	armor_penetration = 10
```
Trilbyspaceclone, line 68: Why are you nerfing the magnum?

code/modules/projectiles/projectile/beams.dm
Dimasw99: > Same AP as the parrent
To make it more readable
> why more damage
The damage value was taken from bullet/light_rifle_257/scrap with standard dealing 22 damage.
A deciding factor about taking ammo is how much damage you can do with each bullet to conserve on inventory space and ammo use. With such low values you would be burning through both space and materials

code/modules/projectiles/projectile/beams.dm
Dimasw99: Adjusted to AP of standard. Was feeling odd about it myself and could potentially be balanced out with reducing the material cost of 9mm laser.

Another fact about it being 15 was that it would have the same performance rating as other larger calibers making it a bit awkward.

code/modules/projectiles/projectile/beams.dm
Dimasw99: It was doing more damage than the standard ammo type that dealt 35 damage.

It went against the theme that laser bullets would otherwise have with other calibers.

code/modules/projectiles/projectile/beams.dm
Dimasw99: Huh, that is interesting. I am curious to know where this 20 damage gets converted in 80.

code/modules/projectiles/projectile/beams.dm
Dimasw99: Adjusted to standard bullet AP to fit the theme

code/modules/projectiles/projectile/beams.dm
Dimasw99: Went into a local test to see if this is true. I spawned in a STS with a laser bullet with its damage set to 16 and one set to 20.
My results are 47.6 damage with 16 dmg. 52 with 20 dmg.
The 80 damage conversation cannot be found.

Trilbyspaceclone:
Prob should have named the PR "Rebalances" rather then buffed, other then that the changes are well thought out enuff

code/modules/projectiles/projectile/beams.dm
Trilbyspaceclone: 1 damage mod with a 1 ap mod with vig of 0
image
having basic vig will e-z do 65+
I stated 80 do to thinking with 1.2 pen guns rather then defaults
and stats

Trilbyspaceclone: Note this is with **SERVER NUMBERS**
15 more ap (that gets dobled when in terms of damage) to that with the extra 4 more burns
can with easy deal 80 damage

Trilbyspaceclone (approved):

Trilbyspaceclone:
After more talking please doble make sure this is ready

Dimasw99:
With my recent discovery of how scuffed bullet code works with VIG affecting it I kinda need to put this PR on hold as I will have to create a proper spreadsheet for it as the information presented is miss leading.

Funnily enough it still makes people believe that laser bullets are shit just because the gun info telling them so. Probably something that should be adjusted

PR #3670, opened 2022-07-14
Medical Perk-Checking

Stutternov:
## About The Pull Request
Adds code to check for Medical's surgery perk (Or herbalist for hunters) to do surgeries or a bio of 50+. Some surgeries are locked behind this - mostly relate to organ repair, bone grafting and other surgery types. This 'lock' can be bypassed by using Ultrasurgeon, granting a temporary perk atop of its increase in biology allowing you to attempt surgery. Shrapnel removal has seen a **slight** increase in difficulty - but still has a 100% success rate if the person has 40 bio skill basically. Self surgery has also been penalized **heavily** by having a doubled fail-rate from what it was and a doubled time-wait.

As every 'balancing' PR is a take-and-give, I've lowered the requirement to use the advanced kits **heavily**. People who are non-Soteria with a bio skill of 25 (instead of 75) can now use it. This may be adjusted but I felt it's fair since the Absolute ones are 15 bio skill and syringes are also 15 bio skill.

This should make Soteria's job more focused on as it requires people to either use splints to temporarily reduce an injuries' penalty or to go to Soteria. This is mostly a start of my intentions, as I believe allowing players to freely heal themselves of non-major injuries and requiring Soteria or a heavy stat-bypass for more serious injuries. 

## Changelog
:cl:
add: Stat/Perk 'locks' for certain surgery types. Only really involves 'complex' surgeries that are not life-saving or simple.
tweak: Edits shrapnel surgery to actually have a tad bit of low-risk 'randomness' to it to encourage people with actual biology skills to do it.
tweak: (?) No idea if it was broken by code-line, but added a new check to prevent surgery on splinted limbs. Just remove the splint though and you can do surgery as normal.
balance: Allows advanced kits to be used again by those of at least 25 bio. (Subject possibly to reduction if needed.) 
balance: Reduces biology needed by syringes by 5, allowing someone with a bio of 10 to now use syringes.
balance: TT has received 5 extra bio skill since it took Orderly's bio skill and not paramedic. To prevent heavy stat-powergame at round start to bypass the checks, 5 biology has been removed from Corpsman and Salvager.
balance: Because people enjoyed doing self-surgery to avoid medical fees, the difficulty and time of self-surgery has been doubled! Yipieeeeee!!
/:cl:

### DIFF
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -5,10 +5,13 @@
 	var/list/allowed_tools = null
 	var/required_tool_quality = null
 	var/target_organ_type = /obj/item/organ/external
+	var/perk_i_need = PERK_SURGICAL_MASTER
+	var/perk_i_need_alt = PERK_MASTER_HERBALIST
 
 	var/difficulty = FAILCHANCE_HARD
 	var/required_stat = STAT_BIO
 	var/duration = 60
+	var/requires_perk = FALSE
 
 	// Can the step transfer germs?
 	var/can_infect = FALSE
@@ -88,6 +91,11 @@
 /obj/item/organ/proc/try_surgery_step(step_type, mob/living/user, obj/item/tool, target, no_tool_message = FALSE)
 	var/datum/surgery_step/S = GLOB.surgery_steps[step_type]
 
+	if(S.requires_perk)
+		if(!(user.stats.getPerk(S.perk_i_need) || user.stats.getPerk(S.perk_i_need_alt)) || !user.stats.getStat(STAT_BIO) >= 120)
+			to_chat(user, SPAN_WARNING("You do not have the training necessessary to do this surgery!"))
+			return FALSE
+
 	if(!S.is_valid_target(src, target))
 		SSnano.update_uis(src)
 		return FALSE
@@ -128,14 +136,14 @@
 
 	// Self-surgery increases failure chance
 	if(owner && user == owner)
-		difficulty_adjust = 60
-		time_adjust = 20
+		difficulty_adjust = 120
+		time_adjust = 40
 
 		// ...unless you are a carrion
 		// It makes sense that carrions have a way of making their flesh cooperate
 		if(is_carrion(user))
-			difficulty_adjust = -150
-			time_adjust = -40
+			difficulty_adjust = -300
+			time_adjust = -80
 
 	if(user.stats.getPerk(PERK_ROBOTICS_EXPERT) && S.is_robotic)
 		difficulty_adjust = -90
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -97,6 +97,10 @@
 			to_chat(user, SPAN_WARNING("You do not have the training necessessary to do this surgery!"))
 			return FALSE
 
+	if(status & ORGAN_SPLINTED)
+		to_chat(user, SPAN_WARNING("You need to remove the brace first!"))
+		return FALSE
+
 	if(!S.is_valid_target(src, target))
 		SSnano.update_uis(src)
 		return FALSE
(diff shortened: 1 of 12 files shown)


cdb-is-not-good:
Based. Good changes, i assume this is all fully tested?

Stutternov:
> Based. Good changes, i assume this is all fully tested?

Yes, although I have a change I am making now. Players expressed issue with the high bio check, so, instead I got pitched a few solutions. One I thought was good was one regarding ultra-surgeon giving a bypass perk to allow surgery at the cost of a bit of loss to the biological bonus ultra surgeon gives. That way while someone can do the surgery needed, such as low-pop revival, they risk higher fail rates. 

Basically making it a 'downside' to people in the field and not much of a benefit but good for emergency treatment situations.

Stutternov:
PR is set now with the changes. Me rn.
https://www.youtube.com/watch?v=PtK_crFV04s

code/modules/surgery/surgery.dm
```diff
@@ -88,6 +92,11 @@
 /obj/item/organ/proc/try_surgery_step(step_type, mob/living/user, obj/item/tool, target, no_tool_message = FALSE)
 	var/datum/surgery_step/S = GLOB.surgery_steps[step_type]
 
+	if(S.requires_perk)
+		if(!(user.stats.getPerk(S.perk_i_need) || user.stats.getPerk(S.perk_i_need_alt) || user.stats.getPerk(S.perk_drug)) || !user.stats.getStat(STAT_BIO) >= 50)
+			to_chat(user, SPAN_WARNING("You do not have the training necessessary to do this surgery!"))
```
ThePainkiller, line 97:
```suggestion
			to_chat(user, SPAN_WARNING("You do not have the necessary training to do this surgery!"))
```

ThePainkiller (changes requested):
Typo

ThePainkiller:
Leaving my two cents as medical main even if they'll get ignored:

25 BIO is too easily accessible by anyone, prospies can get that on two level ups very easily. Considering doctors have 40-ish BIO on roundstart, that number ought to be harder to keep these exclusive to actual medical "professionals". I believe 50 BIO is a good middle ground for them. 15 BIO bypass on church kits is laughable, honestly. Why do they even have such low requisites while being the most brute-healing-per-application pack in the game is anyone's guess.

Self surgery penalties were decent enough that even with a hundred BIO I've botched them so far as to jump sensors more than once. Those numbers are a ridiculously high jump to a downright impossible level unless you're a hundred plus BIO. By this point just outright removing self surgery from being a thing does literally the same effect, since you'll have to be four hours into a shift of pure BIO artifact minmaxing leveling to even be able to.

Stutternov:
> Leaving my two cents as medical main even if they'll get ignored:
> 
> 25 BIO is too easily accessible by anyone, prospies can get that on two level ups very easily. Considering doctors have 40-ish BIO on roundstart, that number ought to be harder to keep these exclusive to actual medical "professionals". I believe 50 BIO is a good middle ground for them. 15 BIO bypass on church kits is laughable, honestly. Why do they even have such low requisites while being the most brute-healing-per-application pack in the game is anyone's guess.
> 
> Self surgery penalties were decent enough that even with a hundred BIO I've botched them so far as to jump sensors more than once. Those numbers are a ridiculously high jump to a downright impossible level unless you're a hundred plus BIO. By this point just outright removing self surgery from being a thing does literally the same effect, since you'll have to be four hours into a shift of pure BIO artifact minmaxing leveling to even be able to.

If staff prefer I'd be 100% in favor of removing self surgery as it is, the only reason I didn't is due to FBPs and tool upgrade ability to negate downsides of self-surgery. 

I've only seen it used by players doing surgery on themselves in the field and using certain species who do not feel pain or hyping up on painkillers for it.

Stutternov:
> I feel like 25 bio for ATKs is honestly fine, particularly given that while it will allow even a relative amateur to clear serious burns/brute they still won't be able to fix organs, bones, brains, etc. At least, if i understand this right.

Correct. You'd be limited by the surgery's bio-requirement of 50 minimum or the bypass perks. To use the ATK kit in surgery it requires 50 bio default. Funnily I didn't even need to change that, it was already like that.

Stutternov:
Added a small update fixing a possible error/oversight (?) I saw in other code lines of splints supposed to make stuff unable to do surgery past. Not a big change, really mild, but I guess it fixes that since I got it working.

Simply makes splinted limbs require the splint to be removed to operate on the limb. So, same as usual, just remove the splint to do surgery on it.

cdb-is-not-good:
Unless someone offers a complaint against this PR i'll likely merge this tomorrow

PR #3706, opened 2022-07-23
Empty linked box on teleporter

ThePainkiller:
## About The Pull Request

I added the Heroic to the teleporter, but there's no way to use it if you don't have the right mag (or in this case, linked box) to load it with, unless the armory is open (which means you could instead use the one in the armory) or printed through Guild. This PR adds an **empty** linked 7.62mm ammo box for the Heroic (and the Pulemyot) on the teleporter for 65 power.
	
<hr>

## Changelog
:cl:
add: Adds an EMPTY linked ammo box of 7.62mm to the Blackshield teleporter for 65 power cost.
/:cl:


cdb-is-not-good:
Yea sure this works.

PR #3709, opened 2022-07-23
I hate icon code!I hate icon code!I hate icon code!

cdb-is-not-good:
Spatial cutters sheaths only hold spatial cutters now.

Small pathing fix for the medical belt, now properly displays on-mob sprite.

Fix for pretty much all belt-sheaths, when worn on the belt they now properly put together their sprite based on which sheath you're using and which sword is in it. Also makes the belt sheaths wearable on the suit slot of most armors, a tradeoff for the fact that they cannot be attached like standard sheaths

There MAY be some that I missed, let me know if that's the case :)

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


cdb-is-not-good:
Also know bug but one that I'm not going to fuck with for now, when worn on the suit slot the sprite won't update on its own. Oh well.

PR #3777, opened 2022-08-05
Marshal Bullet Fab, Prospector Scrap Crafting & Misc Changes

Stutternov:
## About The Pull Request
After talking the previous night with a few people and receiving suggestions, I've redone the Marshal's armory a little bit.
- The Marshal autolathe has been moved out into the Marshal aux-storage, removing a few barricades to make room for it.
- Barricades removed have been given to Blackshield's armory which has received a slight expansion. They use them more than Marshals do anyhow.
- Marshal armory has received a bullet fabricator; this is an alternative to the scrap ammo kit creator we had, as we never used it.
- Prospectors have received the scrap ammo kit printer **as they were supposed to be rewarded months ago with new gear.** So - here, you get the scrap kit maker since I'd think prospectors may use it more.

[hr]
Marshal storage:
image

Marshal armory:
image

Blackshield armory:
image

Prospector prep: (Note - I did keep both random disks spawns. They're just stacked now.)
image

## Changelog
:cl:
added: Bullet fabricator to Marshals. 
added: Scrap kit creator machine to Prospectors.
balance: Made other various misc changes such as moving stuff around.
/:cl:


cdb-is-not-good:
I would strongly prefer to split the bullet fab as previously discussed but neverminding that. Please do me a huge favor and put proper armory access on the armory windoors

Medizeal22:
The Mapping Gods have Spoken and they approve for mapping.

ThePainkiller:
I know a lot of guild players would complain, but the Bullet Fabricator made more sense for Marshals to have than them in all honesty....

cdb-is-not-good:
I agree but I have issue with just duplicating the fab outright

ThePainkiller:
What I meant with my comment is that the fab on guild should be moved to marshals, not having two of the same machine on two different departments :P Guild should get something else to help compensate their loss though.

cdb-is-not-good:
Agree. The original discussion tk which I referred would split the fab into two- one with vanillas fmj rounds and one with exotics- such that only say....the people responsible for handing out explosives could produce them :p

cdb-is-not-good:
ehhhhhh yea fuck it, if we want to adjust the balancing later or maybe take the one from the guild or split them or whatever we might want to do it can get decided down the line. This PR has enough stamps at this point that i dont wanna leave it waiting any longer.

PR #3812, opened 2022-08-09
"Optimizes" superior mobcode with somewhat of a nuclear option in terms of optimization. Fixes stealth bird targetting, fixes mobs being permastunned as well as not getting up from knockdowns. 

nikothedude:
## About The Pull Request
I am not fucking happy about this, but by god I can not figure out a better way to do this. Get regular status updates is a big cause of superior mob lag, rivalling findtargets at times. Since most mobs dont actually really care about reagents much, which to my understanding is the only thing that doesnt directly force a health update (WHY IS IT ONLY REAGENTS), this should cut the cost of this proc by 66%, in turn further reducing the cost of superior mobs. 

I however had to replace the cheap lying and buckled verb status proc with a more expensive one from living. This is a much better proc as it actually knows when the fuck the mob is getting up. This is the fix for mobs never getting up. 

Some tesitng has shown, however, that even with this, get regular status is still more expensive than cheap lying and buckled verb, /even though its on LIVING and not SUPERIOR./

Is this worth it? I don't know. It might be. 

## Changelog
:cl:
fix: Mobs now get up after being stunned.
fix: Invisible vox now properly target.
code: Superior mobs now only update their health once per 3 ticks.
/:cl:


nikothedude:
hoyl fuck. this avenue of optimization is impossible until mobcode is rewritten good fuck

cdb-is-not-good:
If it works and doesn't break anything then I suppose we'll see if the cost is worth it on live

PR #3838, opened 2022-08-12
Aberrant Organs - Organ fab update, RNG reduction, balance pass

Pink-Chink:
### Updates

- Updated the organ fabricator UI and added design disk ripping for Oh My Guts! disks. Also, it can be upgraded, but deconstructing the machine has undesirable results (player is given a warning beforehand). Use a part exchanger to safely upgrade.
- Teratomas are now separated by organoid type. There is still randomness in the exact input or output, but this should greatly reduce the time and resources needed to produce a viable organ. Disks are updated with the new types.
- Premium inventory on organ vendors moved to contraband.

image

### Balancing

- Minor buffs to organs in OMG! The Classics vendor and related disk.
- Nociceptive organoids no longer have halloss as an option. It's too much of a catch-all and makes the other types less relevant.
- Healing chems were moved from the standard pool of reagent outputs to the uncommon and rare pools.
- Lowered max organ efficiency from a generated organoid to 25 from 33.

## Testing

	Print from design disk
	Rip OMG! design disk
	Rip Back Alley Organs
	Replaced organ fab parts with RPE
	Deconstructed organ fab
	Opened deconstruct warning window and moved away from fab before hitting deconstruct
	Attempted to apply an input organoid on an organ with an input organoid installed
	Printed a teratoma of each type

## Changelog
:cl:
add: Updated UI for organ fab and added OMG! design disk piracy
tweak: Biomass spillage when organ fab is deconstructed with biomatter sheets inside
tweak: Organ fab can be upgraded
tweak: Teratomas are separated by organoid type instead of input/process/output
tweak: Renamed "superior endocrinal organoid" to "intracrinal organoid"
tweak: Added categories to organ designs
balance: Buffed organs in OMG! The Classics vendors and related disk
balance: Moved standard organ functions in process organoids to uncommon variants
balance: Removed halloss from nociceptive organoids
balance: Moved healing chems in gastric/hepatic organoids to uncommon/rare variants
balance: Lowered max organ efficiency from a generated organoid to 25 from 33
fix: Fixed issue with multiple input/output organoids on one organ
fix: Ruined teratomas properly lose their upgrade slots
/:cl:


Pink-Chink:
I'm not well-versed in Soj's mechanics since I'm porting this from my own Eris downstream. People have been asking about a way to produce organs without needing to rely on the church. So, it's definitely something I'll look into.

nikothedude:
Btw there's a new contributor guideline to list the steps that you took to test your PR, so you might want to do that

PR #3846, opened 2022-08-13
Kriosan mobs for the upcoming event

Stutternov:
## About The Pull Request
<details>
<summary>
	As title says. Requested by Lich.
</summary>
</details>

## Changelog
:cl:
adds: Event mobs for the event.
/:cl:


nikothedude:
don't forget to set comfy_distance, comfy_range - comfy_distance is how far away from a target mobs will stop moving at, comfy_range is the firing range

code/modules/mob/living/carbon/superior_animal/human/subtype/kriosan/kriosan.dm
```diff
@@ -0,0 +1,93 @@
+/mob/living/carbon/superior_animal/human/rebel
```
nikothedude, line 1: also this is just personal preference but i usually don't like using base types in any capacity that isn't for holding vars, it makes things messy when i want to make a change to this specific type but none else

cdb-is-not-good:
I assume these are the specifications to which lich wanted these mobs made?

PR #3847, opened 2022-08-13
Blob Spawning Rework #1

ApplePlastic:
## About The Pull Request
<details>
<summary>
	Prevents blob from spawning in the main R&D lab and Robotics, which are both critical areas. 
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

### DIFF
--- a/maps/__Nadezhda/area/_Nadezhda_areas.dm
+++ b/maps/__Nadezhda/area/_Nadezhda_areas.dm
@@ -1457,6 +1457,7 @@ area/nadezhda/medical/medbaymeeting
 /area/nadezhda/rnd/research
 	name = "\improper Research and Development"
 	icon_state = "research"
+	flags = AREA_FLAG_CRITICAL
 
 /area/nadezhda/rnd/researchbathroom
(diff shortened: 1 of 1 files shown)


cdb-is-not-good:
Why are they critical areas?

ApplePlastic:
The blob can delete everything in the R&D lab (autolathe, circuit printer and consoles included) and make it a hinderance for the rest of the round. Most things would have already been eaten due to the blob announcement's delay, since it only alarms when it is of considerable mass.

cdb-is-not-good:
Doesn't the cro have a disk with all the stuff to make a new lab?

ApplePlastic:
I am not entirely sure on that, honestly. This only prevents the blob from spawning directly inside these areas- anywhere around it is free game, if it spreads there fast enough. 

You could spend an hour or two gaining materials for the department, just for them to be deleted instantly by the blob happening to spawn right next to the autolathe (which is what happened in a very recent round).

cdb-is-not-good:
Fair and reasonable

PR #3899, opened 2022-08-25
Henryk and Two-Face iconic suits 

ApplePlastic:
i love henryk and two-face

## About The Pull Request
<details>
<summary>
	Adds Henryk Niebieski's suit, as well as Two-face's black and white suit. 
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
Adds the teal and two-tone suit to the loadout. 
/:cl:

PR #3902, opened 2022-08-25
quick fix to my suits

ApplePlastic:
changelog not needed, shit's a hotfix to #3899 (i went on the server and i couldn't select them in the loadout!!!)


ApplePlastic:
why is this conflicting with mapdiff........

PR #3903, opened 2022-08-25
fixes suits from recent PR 

ApplePlastic:
changelog not needed, supposedly fixes #3899 (it conflicted with mapdiff for some reason)


ApplePlastic:
why does it 
conflict with mapdiff

cdb-is-not-good:
It doesn't. The conflict is in code/modules/clothing/under/suits.dm. Because you tried to make a PR on the same branch without making a new  one. Basically your PR is trying to make the same change twice, which is bad. Always make a new fork when making new PRs

ApplePlastic:
real

PR #3920, opened 2022-08-29
Makes the Akula tail fit non-Akula bodies.

LiteralMushroom:
## About The Pull Request
<details>
<summary>
I literally just added a few more pixels to the back of the Akula tail, for people who want the tail WITHOUT a shark head.
Not tested yet.

I might've added too many pixels but I don't think that matters since tails go behind your body anyways.

edit: nevermind they don't and therefore closing this because I can't make it fit Akula AND everyone else and I don't want to make a separate tail

edit: nevermind AGAIN apparently it's supposed to look like that, cool
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
fix: Akula tail has a few more pixels so now it actually works on bodies other than the shark one.
/:cl:

PR #3945, opened 2022-09-08
Revert "Makes the Akula tail fit non-Akula bodies." because Sojourn doesn't have a functioning layer system.

LiteralMushroom:
Reverts sojourn-13/sojourn-station#3920

the tail fucking goes over clothes
why do we not have a functioning layer system???

Why did I not realize this before the original PR got merged? Because I was dumb and only looked at the naked sprite.


cdb-is-not-good:
there's a switch tail layer verb. Did you try that first?

LiteralMushroom:
> there's a switch tail layer verb. Did you try that first?

it
goes over clothes in character setup

cdb-is-not-good:
> > there's a switch tail layer verb. Did you try that first?
> 
> it goes over clothes in character setup

That doesn't really answer my question. Got a picture of the thing?

ThePainkiller:
> there's a switch tail layer verb. Did you try that first?

The switch tail layer verb does not work as it was poorly ported without taking into consideration that it applies to ORGANS and not character markings, which are two different things. I can attest to it never having worked, ever.

cdb-is-not-good:
> > there's a switch tail layer verb. Did you try that first?
> 
> The switch tail layer verb does not work as it was poorly ported without taking into consideration that it applies to ORGANS and not character markings, which are two different things. I can attest to it never having worked, ever.

weird- it works for me but admittedly it does very little other than make the tail appear over the backpack layer which ...barely matters.

LiteralMushroom:
> > > there's a switch tail layer verb. Did you try that first?
> > 
> > 
> > it goes over clothes in character setup
> 
> That doesn't really answer my question. Got a picture of the thing?

Capture

significantly less noticeable if you're naked
which is why I didn't realize it was a Problem until now

Trilbyspaceclone (approved):

PR #4125, opened 2022-11-12
Sheet Snatchers Offers!! Wow!!

WilsonWeave:
Makes Lancer station offer 300 credits for up to 4 sheet snatchers at a time, considering Solnishko buys many more knives at a time, for 150 each, this seems more than fair considering the effort  and materials involved. Also ain't no one cooking food for trade offers chief.

Also makes Boris station offer 400 credits for up to 2  guild made advanced sheet snatchers at a time.  Having literally only one, extremely niche and rare to obtain offer as the only offer for a station is a really bad idea. Might not make the most sense for this station, but I'll consider replacing it with something more fitting. Eventually. But it works for now.

...Also it's been awhile since I've made a PR so yell at me if I fucked it up somehow.

And, tested to make sure it works and compiles 

image

image

### DIFF
--- a/code/modules/trade/datums/trade_stations_presets/2-uncommon/rigprovider.dm
+++ b/code/modules/trade/datums/trade_stations_presets/2-uncommon/rigprovider.dm
@@ -38,5 +38,6 @@
 	)
 	offer_types = list(
 		// TODO: offers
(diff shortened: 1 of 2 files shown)


Stutternov:
wow!!! thank you, wilson!!! speed merge this!!!

PR #4129, opened 2022-11-13
Removes bone modding (bug fix)

LongSleeves:
Modding bones caused their HP to vanish and potentially "bricked" people. Bone mods are disabled for now.


Pink-Chink:
Muh bones

PR #4134, opened 2022-11-13
Caliber Damages & Recoil

Stutternov:
## About The Pull Request
Simply - this PR does these few things:
- Ups the effective range of 9mm to not instantly lose its drop off after traveling a single tile from the user firing it.
- Ups the recoil on 12mm / .50 Kurz. The reason being - the rounds are more powerful than even a 7.62 rifle. As such, they deserve a higher recoil. AP is good to have, but with the changes of AP damage to mobs, it is not as powerful as it once was.
- Ups the damage but also the recoil of 8.6mm / .408. The reason the same as above - its AP is strong, but its damage is 'meh'. This should make .408s act as proper hard-hitting battle rifles but hard to fire in large bursts of full-auto.
- Increased some agony damages on HPs to make them more viable. Also upped some of their recoils  - no idea why they were lower.
- 6.5mm has received a slight recoil increase. Why did 6.5mm carbine / rifle calibers have a somehow lower recoil than a 9mm handgun despite a larger casing and velocity?

## Changelog
:cl:
balance: Ups recoil on 12mm and 8.6mm.
balance: Ups effective range on 9mm.
balance: Ups damage on 8.6mm.
balance: Ups some agony damage and some recoil on agony damage bullets.
balance: Ups the recoil on 6.5mm rounds to not make them somehow lower than 9mm.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullettypes.dm
+++ b/code/modules/projectiles/projectile/bullettypes.dm
@@ -13,8 +13,8 @@
 	step_delay = 0.65
 	can_ricochet = TRUE
 
-	affective_damage_range = 2
-	affective_ap_range = 2
+	affective_damage_range = 3
+	affective_ap_range = 3
 	recoil = 3
 	added_damage_bullet_pve = 4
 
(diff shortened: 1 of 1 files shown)


Lamasmaster:
bidon

PR #4138, opened 2022-11-14
Basic Chemist Offers, Combination Driver Slash. Famine.

WilsonWeave:
Sets non-default prices for meat, tofu, eggs, flour, and milk for Vermouth station.

Cad station has gotten a set of basic medicine bottles as offers, to give something chemists to do quickly and easily. Or the church/garden that knows how to work with plants.

Gives Dylovene a non-default price for the SI station.

Decreases the value of both normal and Greyson combination drills, and ups the required materials to print both.


WilsonWeave:
image
image

WilsonWeave:
I did a silly and deleted this one instead of the branch I meant to

PR #4158, opened 2022-11-18
Nanogate Balancing

Stutternov:
## About The Pull Request
<details>
<summary>
	Balanced out nanogates finally. Nanogates now actually have IMPACTS from being EMPed and damaged. When the Nanogate takes and EMP, depending on its severity, it does temporary agony damage to the user. If the Nanogate goes below 10 health, it will break, causing the user intense looping pain and organ damage to either their heart, eyes, or brain, until the implant is removed or fixed. This will **FINALLY** provide a proper downside to Nanogate; as genetics is rarely used with a nanogate anyhow and the inspiration is honestly negligible due to perks you can get from 1 inspiration usage.
</summary>
<hr>
<hr>
</details>

## Changelog
:cl:
balance: Finally adds an effect for Nanogates when they get damaged rather than being funni meme: "YOU HATH NO EFFECT ON ME!!!" when they are damaged via an EMP or brain-trauma.
/:cl:

### DIFF
--- a/code/modules/organs/internal/nanogate.dm
+++ b/code/modules/organs/internal/nanogate.dm
@@ -23,6 +23,7 @@
 	var/mob/living/carbon/superior_animal/nanobot/Stand // The personal robot of the owner. I wonder how many people will get the reference... -R4d6
 	var/obj/item/rig/nanite/nanite_rig // The nanite rig you can make
 	var/list/perk_list = list() //List of activated perks for later removal
+	min_broken_damage = 10 //Should break when organ is at 10 health of its 60.
 
 	owner_verbs = list(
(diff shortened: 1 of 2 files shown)


ThePainkiller:
Is it possible to make Nanogate MSOF visible on health scanners? Doctors will sometimes outright ignore a broken nanogate because of low MEC to repair it with, and one time even though I repaired the nanogate it kept rotting every organ inside the patient until they woke up and yelled "IT'S NANOGATE POISONING TAKE IT OUT" before dying again, the issue WAS solved when I took it out, revived, and put it back in, but we need a more intuitive way to know this

PR #4159, opened 2022-11-18
No More Infinite Perks from Oddities

Stutternov:
## About The Pull Request
<details>
<summary>
	This PR does one simple thing - effectively reverting the code that made oddities keep their perk after usage. This was a change Eris did last year as well that I still feel is not exactly well balanced and makes oddity perks a 'pass-around' thing instead of encouraging the buying and/or selling of oddity perks. Similarly, some oddity perks are rather strong perks (i.e - Gunslinger, Springheel, etc) and therefor should be limited and treated as rare commodities rather than something to be freely passed around as: "I got my use out of it, here, on the house." _How will the Oddity economy ever recover from this mean rebel-man???_
</summary>
<hr>
<hr>
</details>

## Changelog
:cl:
tweak: Tweaks oddities to be a 1-time use in terms of perks. After the use, the perk disappears but the oddity will keep its stats.
/:cl:

### DIFF
--- a/code/modules/sanity/sanity_mob.dm
+++ b/code/modules/sanity/sanity_mob.dm
@@ -280,7 +280,8 @@
 			to_chat(owner, SPAN_NOTICE("Your [stat] stat goes up by [stat_up]"))
 			owner.stats.changeStat(stat, stat_up)
 		if(I.perk)
-			owner.stats.addPerk(I.perk)
+			if(owner.stats.addPerk(I.perk))
+				I.perk = null
 		for(var/mob/living/carbon/human/H in viewers(owner))
 			LEGACY_SEND_SIGNAL(H, COMSIG_HUMAN_ODDITY_LEVEL_UP, owner, O)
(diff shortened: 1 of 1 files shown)


ThePainkiller:
This hurts, but I heard long ago we were intended to have this feature as soon as Eris got it, so fair

Stutternov:
> This hurts, but I heard long ago we were intended to have this feature as soon as Eris got it, so fair

Ngl I'd sooner limit usages on oddities in general, i.e 2 usages and the oddity destroys itself forcing people to get a new oddity.

But I'm lazy, that would be way more difficult to code, and oddities are candy since they can be made/produced - and no one has rebalanced that. 

My main 'fear' is that prospectors will never sell oddities with good perks on them and instead use them themselves - but, to be fair, **they already do that** by passing perks around and never selling it 'incase another prospector comes'. Sadly we aren't Eris with vagabonds not being a department being the main source of oddity-finding, so sales of oddities are likely lower. More-so because 'credits are worthless'.

PR #4219, opened 2022-12-10
Wetfloor Visual Fix for the blind and speedy

Dongels:
**Description:** _"Mostly a simple tile effect change to help those speedy colonist avoid getting their face slapped on the wet floor as often, we all suffer in this purgatory."_ 

**What was changed:**  =>  Cut the animation time in half to help with being able to notice that the floor IS WET.
**Was this tested?:** =>  No mostly due to it only being animation timing change (If this breaks the server I'll be amazed with my POWAH)


Dongels:
I would like to request this also be labeled as a "Salt PR" as I do not wish to seem like I and lying to anyone here, or the 10% salt content of the water I have been graced to lay my developer hands upon this pr. *bow*

PR #4329, opened 2023-01-26
Enclave Bot Salt

Stutternov:
## About The Pull Request
We increased explosive damage to cause death so easily due to something as simple as bleed damage - which even rough life cannot prevent.

Mob should not be density of 0 when it can nearly instant-kill the target and, if they somehow survive, will bleed out 70% of the time.

## Changelog
:cl:
balance: fixes the Enclave-eyebot rip to be less broken.
/:cl:


OliOliOnsiPree:
Enclave badman

PR #4358, opened 2023-02-02
Giving blackshield an anti-armour weapon in their armoury. 

Chevalt:
## About The Pull Request
<details>
<summary>
	Gives blackshield a viable anti-mech weapon with EMP rounds, because they're the colony's militia and generally are the ones to deal with heavier threats. 
(Post Scriptum, ignore the commits that do not affect the Omni_STU map, they do not change anything and are a result of another outdated, unused map having unknown items)
</summary>


Stutternov:
IMO - I'm fine with Shield getting an RPG personally but they should not get more than 2 round start rockets. They have a full box of SLAP ammo, grenades, and tons of guns. (Mechs unironically will die faster to SLAP than regular rockets anyway due to the drivers taking damage from them plus rate of fire.) But they should prob lose their tear gas and flashbang grenades or something as trade. I don't inherently have a problem with this since Blackshield, at one point, was supposed to have a China Lake but I have no idea what happened to that.

Rockets should be added to their ammo-disk (non-teleporter). EMP rockets should **NOT** be used - they're stupidly powerful and legitimately just a regular rocket on steroids (90 damage to person hit + blast damage + emp effect). It's why I've not added them to any department or faction but mainly added them with the purpose of antags.

Chevalt:
You're free to add the rockets on an ammo disk, but I thought of giving them a few because they are in EXTREMELY limited supply across the entire map. They cannot be bought nor produced, so to me, it's balanced, considering the RPG has much less DPS than most other weapons.

Stutternov:
> You're free to add the rockets on an ammo disk, but I thought of giving them a few because they are in EXTREMELY limited supply across the entire map. They cannot be bought nor produced, so to me, it's balanced, considering the RPG has much less DPS than most other weapons.

I would just prefer they _work_ for the extra ammo for them. I actually don't even mind if they produce rockets, Ironhammer on Eris code are _actually_ supposed to be able to - or anyone with the Guild craft station. As such I'd say to give them a limited supply of rockets and make them producable.

I could make them producible on my gun-parts PR that's up rn since I'm porting Eris' gun-station code for Guild anyway. So either could have them craftable that way by Guilld, making Shield require some trading - or could add them to their ammo disk at a high cost so they have a reason to use it.

I'm by not means the dev-head but I'd **heavily** suggest against the EMP RPGs for the reason stated, I coded them in I _think_ right before or after the Excelsior base attack. I love the things, but they're way too powerful and will kill _anything_, it's just a 'point and win' rocket. I'd would personally prefer if they were kept as antag-only just so antags can have a 1-up on the colony due to always being outnumbered.

Chevalt:
We can merge the PR and then modify the starting armament of the blackshield once your own PR that allows building rockets is done, I'll nerf a bit the EMP rockets in a separate PR because it makes sense from a balancing perspective (specialised EMP munitions are bound to have less HE filler than standard HE rockets, it's obvious, so I'll implement it)

PR #4362, opened 2023-02-03
Adds Female sprites for the church robe

KitchiFox:
adding the female version as it completely slipped my mind.

Women now may wear the robes.


KitchiFox:
Made a woopsie. fixing it.

PR #4375, opened 2023-02-04
Adds more Gun Frames, Parts, and Items to LSS Beacon & Armory

Stutternov:
## About The Pull Request
This PR simply adds new disks to the Marshals to encourage gun crafting by the Supply Specialist for Marshal firearms, now having infinite prints on frames and a disk capable, with limited prints, of producing most required parts for Marshals and high-end weaponry. 

Lonestar now has access to gun parts and frames via their beacon, of which they can print a select amount of frames as well via disks they can buy. These disks are expensive but are a return for investment, allowing a print of an estimated 10 guns with the expectation of self-assembly; the guns costing more than their print cost. Lonestar can now act as a 'back-up' for gun parts if Guild is not around or if maints has not rewarded you with enough parts to use in your frame.

Marshal armory has been moved around a bit. Added a new autolathe to the armory itself so that the Specialist can keep their gun frames separate from the public Marshal autolathe. (If this becomes an issue, such as Specs stealing shit from the locker-room lathe, this may be reversed.) Similarly, I've changed around the layout to make more room to move without it feeling so cramped. 

Finally, tons of misc changes regarding recipes, improper crafting, and serial number issues. A follow-up PR is currently being worked on regarding the Guild crafting station and maint-loot gun frame spawning being low.

Thank you for your patience and please report any issues or errors to us on the Discord's bug reports or development channels.

## Changelog
:cl:
add: LSS beacon now can buy parts and gun frame collection disks from 3 vendors.
add: Marshals now properly get their gun frames and parts on a disk. Parts disk for Marshals is limited, frames are unlimited.
remove: Marshal negotiator disk now has most of its designs transferred to frames disk. Contains only specialty items now.
fixes: Missing heavy barrel barrel sprite.
fixes: Issue with Freedom SMG not having a working frame.
fixes: Colt giving wrong crafting result.
fixes: AK frame having Excel AK serial number
fixes: Various other minor inconsistencies with craft results v parts used.
/:cl:


Rat-Time:
Not sure if it was addressed but outsiders spawn with a frame instead of a gun now.

PR #4385, opened 2023-02-06
The Great PvE Damage Removal

Stutternov:
## About The Pull Request
This PR removes **all** mentions and uses of PvE damage modifiers. The PvE damage system was an attempt at a change for the better; making weapons less deadly vs players while keeping them at their pre-nerf values against mobs. This is a sensible change but has unintended consequences. I will list some issues with it below:

- Larger caliber bullets effectively remains incredibly competitive, while lower-caliber rounds suffer due to this system - sporting low damage vs players and, due to mob health and their new armor, low damage against them as well.
- This has led to a slow power-creep in mob healths while not reviewing old mob-health issues. As we keep adding more and more mobs, we tend to avoid wanting to touch their health. I believe that the PvE damage system was a way of standardizing this damage without needing to touch their health; understandable but it creates problems down the line as we're seeing.
- Gun mods making guns **incredibly** powerful are still, in part, due to PvE damage addition on top of regular damage addition too. While this doesn't make PvP any worse, it incredibly buffs stronger guns in PvE more than they were before.
- Armor over-pen has _stayed_, as it probably should. I have reversed my opinion on this as the code now has changed for how it is calculated and applied. I believe over-pen should makeup for any issues in the reduced PvE damage of bullets.
- The system is hard for new players especially to understand. It's also near impossible to explain why this exists in-character.

And finally - 
- I would argue I _preferred_ bullets being deadly, as did a good number of players. However, this will not be changed for now. But this does allow us to decide if we want to do this at a later time, as it 'resets the slate' in a manner.

## "But Rebel, what does this actually do?"

A **tl;dr** of this PR:
- I've reduced mob healths, nearly by half, across the board. (PvE damage system made guns doe near 2x damage vs mobs, so reductions in health are halved roughly to be fair. This will require more fine-tuning.)
- PvE damage modifiers have been fully **removed**; both off guns, their mods, and ammo.
- Some gun mods have received a few tweaks to balance them. Eris' gun mods are a good basis for how we should honestly be balancing ours.
- 12mm has received extra recoil to try and mimic how an actual .50 AE round would feel, rather than it being just 'a bit heavier than a 6.5 rifle'. Same for 10x24 Caseless, it's a literal rifle caliber with 210 grain / 13.6 gram in lore.

## Changelog
:cl:
removes: PvE bullet/laser damage system and modifiers.
balance: Re-balances mob health to avoid them being TOO much of bullet sponges with this change.
balance: Re-balances a few gun attachments, like the gyro and vibro grip, to sport some downside to them. Also gives a slight bonus to one-handing recoil now.
balance: 12mm is cringe. It has received proper recoil values to mimic a 12mm round; it has remained powerful however.
balance: 10x24 is now properly balanced as a rifle caliber since that is how it is used. In Aliens lore, it's 210 grain - which is near equal to a .308 hollow or .410. It should be treated as such.
/:cl:


DorfGuy1:
hey I think the nerfs to vibrocomp is a bit much

PR #4463, opened 2023-02-20
APC -> TGUI (another Guild buff...really)

Dissentbay:
## About The Pull Request
<details>
<summary>

- Reroutes code from nanoui to tgui, some useless buttons kept, they do nothing since their action is commented out
- I just slapped TGUI template and adjusted the data, button functionality and silicon ifs
- silicons can use the APC just fine, overload lights button still there
- hacking is a separate thing and uses nanoui, but the two talk just fine since they are ui, not the under the hood stuff
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
code: Area Power Controller (APC) are now using TGUI, extra functionalities from TG remain, but do nothing

/:cl:

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -797,7 +797,7 @@
 	if(wiresexposed && !isAI(user))
 		wires.Interact(user)
 
-	return nano_ui_interact(user)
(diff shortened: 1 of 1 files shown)


Dissentbay:
woe is the same TGUI error we keep having 😔

PR #4486, opened 2023-02-24
I don't have an addiciton to mech games totally

Lamasmaster:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

After my last PR a realization set in that most people can't actually make sense of the designation I gave to weapons and all so this should make it bit more clear on what is what on top of making mech weapons actual weapons not having a machine gun that has effective damage and range of a scrap pistol

**Ballistics**
Ancient Autocannon is now an Industrial Coilgun, trading it's precision for speed and making it way more clear it's not an actual military grade weapon

Flak Cannon got a reverse treatement, devastating in melee range but borderline impossible to hit at distance

Both of these now cannot be mounted onto combat mechs only industrial

Heavy Machine Gun - overall a pure buff, now actually deals damage similar to a handheld machine gun with hampered firing rate

Can be mounted onto both types still

The three autocannons are now more obvious on what they do
Regular Autocannon = high damage, decent firerate, solid range, Low ammo capacity 
Burst-Fire Autocannon (was U-AC) = less damage, higher firerate (in addition to having 2 round burst), better range, highest ammo capacity
Scatter-Shot Autocannon (was LB-X) = middle ground between the two

**Missiles**
SRM is now ATM (doens't make it much clearer but in some way it might) = 3 salvos of 3 rockets in a row before having to be reloaded
MRM removed
"Long Tom" launcher = rare single use bunker buster, eliminates anything in radius of it's explosive, however the missile is extremely slow (1/10th of normal human movement speed) and way larger than most projectiles. Currently doesn't spawn for obvious reasons so just ask admins nicely for it

Both of these only fit combat mechs now

**Lasers**
New snub-nose PPC cannon = Electrical discharge that has localized EMP and electrecutes anyone caught in it, high damage and very slow firerate, spawns in trash

Mining Laser is now distinct visually from regular laser and no longer has 0 fire delay making it way better than the military one

Heavy Laser = Just little bit better

Multi laser changed into Pulse Laser, shoots 3 beams in a succession, crafting recipe changed since it's not 4 lasers anymore

Pulse Cannon bit lower fire delay

Ammo box capacities changed to reflect weapons having different magazines themselves and made bit more clear on what is each meant for. These also now spawn in trash alongside coilgun and flakcannon/PPC as they were all for whatever reason absent and no way to obtain them.

All weapons have a lot meatier sound to make them truly distinct from regular weapons and have seperate sprites for each one of them (don't claim they are any good not a spriter after all)

Repair drone removed from lootpool but still can be made by soteria / uplinked

<hr>
</details>

## Changelog
:cl:
add: Mech weapons
balance: Mech weapons
soundadd: Mech weapon sounds
/:cl:


AysheDaArtist:
when you need Armored Core VI Fires of Rubicon **now**

elgeonmb:
now add fin funnels

PR #4550, opened 2023-03-12
Mapping fixes, church has a measurable powernet, plus a tweak in Guild's recycling room

Dissentbay:
## About The Pull Request
<details>
<summary>
    just fixing a few pet peeves that I saw and aren't an issue yet because people aren't curious enough
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
tweak: lower cargo now gets atmos pipes from lower maints, since it's really close to atmos itself and saves gases going up a Z  to immediately go down again
image
fix: gas turbine can now get supplied from the burn chamber
image
add: church has a powernet sensor, so guild can measure it, not really much else
image
tweak: guild recycling room had the recycler moved one square east, to save time building the recycling center, and a conveyor got added under the outlet to avoid clogs
aaaaaaaaaaaa

/:cl:


Dissentbay:
a few addendums for clarity:

- lower cargo used to get atmos from the guild's recycling room, I just changed it to receive from guild lower through maints to cut gases going though Z levels without need
- the gas turbine was lacking piping from the burn chamber to it, so I fixed it
- the church power sensor is really just to measure it, guild can change the SMES values already through the rcon
- the changes in the recycling room is just to avoid common issues that happen when making the recycling center, now all that is needed pretty much is to add the smelter and pile ripper, take some glasses down because of space, and you're set

Dissentbay:
a pic of the turbine working
image

PR #4617, opened 2023-03-23
Power to the guild

Lamasmaster:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

Removes most generators and ways to get them from the map except the 3 from spider with 2 of them being moved around, logistics no longer can produce generator boards so you are stuck with what guild has or can make from their nanoforge. Yes this means soteria can get atoms with 1 less research so like eh.
Diesel
Prepper

Removes church solars as they don't really generate much power for the tickrate they devour
churchsolar

Also fixes faulty wiring in soteria outpost (but I hate it either way)
	
<hr>
</details>

## Changelog
:cl:
del: Generator Acquisition
del: Church Solars
fix: Faulty Wire
/:cl:


DorfGuy1:
hey uh. This kind of screws up RND's progression. Soteria science doesn't need an L like this. Plus there've been times where soteria has to fix power when there's no guild.

ActuallyALovelyGrave:
@DorfGuy1 Yes, Trilby and Lamasmaster both ignore all other factors regarding of such harsh changes. I don't know what happens right now in the community, but i would love to see that somebody else points out some weird stuff that being merged without any compains, or otherwise we will see another boring 'high-RP' bay build.

du-top:
Great, who needs power in lowpop anyway.

Nep993:
genuinely a kind of dumb change

PR #4618, opened 2023-03-24
medical buff?

cdb-is-not-good:
Raises the stats required across the board for Body Scanner, Sleeper and Chemmaster to STAT_LEVEL_EXPERT(40), in an effort to cut down on non-medical people doing medical stuff. It seems reasonable that if you are using these machines by sheer knowledge alone rather than training, you would need to be a little more knowledgeable than one CPR class.

Why is this good? It will reduce reliance on advanced medical machines by people who aren't medical, we had these stat locks on before but they were so low as to practically be meaningless. For those lacking medical skill, the cryo pod is still able to suffice to clear physical trauma and dylovene is still OP.

Testing: Did not. Three lines changed, three defines shifted from adept to expert.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -115,7 +115,7 @@
 		add_overlay(image(icon, "sleeper-panel"))
 
 /obj/machinery/sleeper/attack_hand(var/mob/user)
-	if(!user.stats?.getPerk(PERK_MEDICAL_EXPERT) && !usr.stat_check(STAT_BIO, STAT_LEVEL_ADEPT) && !usr.stat_check(STAT_COG, 50)) //Are we missing the perk AND to low on bio? Needs bio 25 so cog 50 to bypass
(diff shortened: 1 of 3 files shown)


Hlscientist:
They and medical players have decided this was a good idea to add in the med-discord. +1

Lamasmaster:
Doesn't this make corpsmen be unable to use health scanners inside their own prep and gatehouse ?

ActuallyALovelyGrave:
It would be great to add even higher check for /obj/machinery/sleeper/hyper, since the superior version of the sleeper.

Loldaddy5676:
is it possible to re add the autodoc into the surface clinic since there really isn't any doctors in low pop?

cdb-is-not-good:
> Doesn't this make corpsmen be unable to use health scanners inside their own prep and gatehouse ?

Corpsmen have the bypass



> is it possible to re add the autodoc into the surface clinic since there really isn't any doctors in low pop?

I suppose? I can look at adding that to this PR when I get home

PR #4641, opened 2023-04-10
Chemistry For the Outlander

Nep993:
For too long, outsiders have been kinda shafted when it comes to medical needs. This gives them access to a chemistry dispenser along with prospectors if they so wish at the void wolf medical facility. These were removed so everyone had to rely on soteria for their medicines, but that's not really an option for outsiders.

literally all it does is readd the chemistry dispensers to the void wolves that were removed like over a year ago.

### DIFF
--- a/maps/_DeepForest/map/_Deep_Forest.dmm
+++ b/maps/_DeepForest/map/_Deep_Forest.dmm
@@ -8475,16 +8475,7 @@
 /turf/simulated/floor/tiled/techmaint,
 /area/nadezhda/dungeon/outside/safehouse)
 "jkl" = (
-/obj/structure/table/standard,
-/obj/item/reagent_containers/dropper{
-	pixel_y = -4
-	},
-/obj/item/storage/box/syringes{
-	pixel_x = 3;
-	pixel_y = 3
-	},
-/obj/item/storage/box/beakers,
-/obj/item/reagent_containers/glass/beaker/large,
+/obj/machinery/chemical_dispenser,
 /turf/simulated/floor/tiled/white/brown_perforated,
 /area/nadezhda/dungeon/outside/smuggler_zone_u)
 "jkn" = (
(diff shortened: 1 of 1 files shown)


SkinkHunt:
It might be better to add one board, and one dispenser. Not certain on that though. 
But otherwise I think this is fine.

DorfGuy1:
this is intended for outlanders, but prospectors will steal it most of the time, resulting in them bypassing medical.

benj8560:
good for outsiders but I really would like prospecting to have more reasons to come to medical that's not them being dead/getting FF shrapnel pulled. Only things I ever see them do. Even with these removed their barely smartanuff to come pay chem a visit as is and this would just give them 0 reason to. Ups and downs I guess

Nep993:
As it stands, outsiders are completely shafted in all aspects from power, to medicine, to weaponry, to armor. Their only saving grace is that they don't ascribe to the laws of the colony... unfortunately for them, that doesn't mean much when they lack resources to actually do anything with their freedom. I've personally spent a full hour just sitting in a corner as an outsider waiting for bandages to heal me because I got jumped by an unlucky render or nightmare stalker spawn.

Yes, prospectors might use it... counter to that however is that they already don't go to medical to begin with unless they're already dead... by that time they're beyond helping with chems.

Nep993:
To add to my prior comment however. Prospectors already don't go to medical if they've got a salvager with any amount of biology. They will at most go to the church and ask for health kits, and usually not even that since they just find ointment and bandages while scaving. Shrapnel is their only real concern, and even then not really if they have the previously mentioned salvager medic to just rip the shrap out of them on the spot.

benj8560:
I still feel like prospecting is simply going to rush it unlimited meds means unlimited prospecting. As well as the outsiders it's intended for not giving getting it. Why not consider putting it in some obscure place away from dungeons that would make for a reasonable mission for an outsider but generally not very profitable for prospecting? Little reason for it not to be their first stop given it usually already is.

benj8560:
I don't recall ever seeing that gear in the hands of an outsider back when we had it and it was usually just sitting in prospector storage while they dumped free/cheap meds for shits and giggles cos haha unlimited.

Nep993:
Here, give me a moment. I have an idea.

Nep993:
image
image
image

Changes made, removed the ones in the void wolf base, added a small triage area next to the crashed outsider ship, along with the means to setup an APC, and a generator with a high capacity fuel tank... giving the outsiders a place to live that's there's effectively.

Tell me what you think?

SkinkHunt:
> this is intended for outlanders, but prospectors will steal it most of the time, resulting in them bypassing medical.

There is one in scrap haven too, if anyone ever bothers to get it. (A few, like Blondie, and SLV, make use of it)

SkinkHunt:
We can try this out and see how it goes anyways.

Guidesu:
huh i okay with it honestly rather iffy would be better to TM first

PR #4643, opened 2023-04-13
rebalancing bounty

Chevalt:
The bounty is a lever action rifle, made to fire 20mm shotgun rounds. But, despite being a quite costly, and supposedly mid-tier blackshield weapon, it's actually pretty bad. And that's because its stats are derived from a makeshift weapon that is purposefully bad, and I can't quite get why. This fixes it, and makes it a decent, but still not very good lever action that trades AP for some damage.


SkinkHunt:
Makes sense. Put it in

Guidesu:
I okay with it

PR #4644, opened 2023-04-14
small changes to colony surface

Chevalt:
This pimps out the prospector porch a little bit, removes a door leading into a part of maintenance that was empty (so now it looks more organic), adds money into the lonestar vault (which by the way we should make a SOP change to tell them NOT to use the precious vault money), bolts the door to the lonestar vault, and that's about it. It also adds a couple of trees. Ignore the changes to colony surface surface map and only account those in the new map, because the old one isn't in use anymore and I just accidentaly clicked on it.


SkinkHunt:
Lets gooooooo :D

PR #4655, opened 2023-04-27
Sanity

Guidesu:
Disables the Hysteric breakdown due to just... how horribly it is handled.
Buff Nilhlism 
And Re-activate a mod that deals sanity damage

### DIFF
--- a/code/modules/projectiles/guns/mods/mods.dm
+++ b/code/modules/projectiles/guns/mods/mods.dm
@@ -548,22 +548,6 @@
 	I.req_gun_tags = list(GUN_PROJECTILE)
 	I.gun_loc_tag = GUN_BARREL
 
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/guns/mods/mods.dm
```diff
@@ -588,6 +572,24 @@
 	I.unique_removal = TRUE
 	I.unique_removal_type = GUN_SCOPE
 */
+//We don't use sanity damage and sanity damaging effects, but I'm keeping this commented out in case we some day do.
+// Add psy damage to your weapon
+
+/obj/item/gun_upgrade/mechanism/psionic_catalyst
+	name = "Moebius \"Mastermind\" psionic catalyst"
```
king5327, line 579: Needs a lore update.

PR #4668, opened 2023-05-18
Error in calculation of station time, fix #4666

### LINKED ISSUE #4666: Guest passes expire too early
I had a guest pass that said "It expired at 09:41." But it's only 9:37 in-game. That has happened before, too.
It seems the guest pass is valid the correct time, but the displayed time seems to be off.
Which is weird, because the variable expiration_time seems to be matching the (incorrect) displayed time, not the (correct) valid time. There seems to be an offset of ~1580 ticks that I can't explain right now. It seems to be unrelated to the difference in colony time and round time (when considering the starting hour)

du-top:
## About The Pull Request
<details>
<summary>
	Fix the display of the expiration time in guest passes
</summary>
<hr>
The variable expiration_time is set to world.time + the duration the guest pass is supposed to be valid. That is correct and the guest pass is valid for the given time. But when examining the guest pass, an incorrect expiration time is shown. This leads to "The guest pass expired at XXX", where XXX is two minutes in the future.
The reason is in the display method worldtime2stationtime. It does not consider that there's a difference between the world start time and the round start time. Subtracting the round start time (in round_start_time) fixes this issue.
This change also affects the autopsy scanner and a display at round end showing when an event started.
fixes #4666 

	
<hr>
</details>

## Changelog
:cl: dutop
fix: Fix display of Guest Pass expiration time

/:cl:

### DIFF
--- a/code/__HELPERS/time.dm
+++ b/code/__HELPERS/time.dm
@@ -19,7 +19,7 @@ var/station_date = ""
 var/next_station_date_change = 1 DAYS
 
 #define station_adjusted_time(time) time2text(time + station_time_in_ticks, "hh:mm")
-#define worldtime2stationtime(time) time2text(roundstart_hour HOURS + time, "hh:mm")
+#define worldtime2stationtime(time) time2text(roundstart_hour HOURS + time - round_start_time, "hh:mm")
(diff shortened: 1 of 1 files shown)


SkinkHunt:
Nice fix

PR #4669, opened 2023-05-18
Enable removing other's underwear, comparable to other clothing items.

du-top:
## About The Pull Request
<details>
<summary>
	Enable removing other's underwear, comparable to other clothing items.
</summary>
<hr>
This is not an ERP mechanics change! Possible correlations to my own presumed fetishes are coincidental and not intended! That made it worse, didn't it? My intentions for this change were:
- you can add underwear to another (appropriate) naked mob by just clicking on it. But there is no way to get it back.
- Performing an autopsy (or a full-body-surgery) on a player who will continue wearing their underwear seems unrealistic to me (immershionnn!)
- All clothing can be removed, but not underwear. Seems inconsistent to me.

Yes, it can be used for ERP and non-con RP, but this mechanic won't change anything about IC or OOC laws/rules. An admin-attack-log entry is performed.

In the stripping dialog (drag and drop from other mob to your mob) there are additional entries at the bottom, one for each underwear item, like it is done on Vesta.Bay and probably even Baystation.
image
The process of removing these items has been added. Existing code has been changed, so that a progress is shown, comparable to the removal of other items.
Checks for items preventing the removal are (already) in place, for example uniform and jackboots.

Based on VestaOfOrion/Vesta.Bay (because I already had a branch of that, I expect that Baystation12 would have worked as well). Revision mentioned in commit message.

	
<hr>
</details>

## Changelog
:cl: dutop
add: Enable removing other's underwear, comparable to other clothing items.

/:cl:

### DIFF
--- a/code/modules/clothing/underwear/base.dm
+++ b/code/modules/clothing/underwear/base.dm
@@ -60,7 +60,7 @@
 		return
 	if(user != H)
 		visible_message("<span class='danger'>\The [user] is trying to remove \the [H]'s [name]!</span>")
-		if(!do_after(user, HUMAN_STRIP_DELAY, H, progress = 0))
+		if(!do_after(user, HUMAN_STRIP_DELAY, H, progress = 1))
 			return FALSE
 	. = RemoveUnderwear(user, H)
 	if(. && user != H)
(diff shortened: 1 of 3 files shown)


SkinkHunt:
Rofl, sure. 
Of course if you that would PR this, Dutop

benj8560:
ironically yes btw. Bay has this yet any ERP sever not directly downstream from this as far as I'm aware is missing this. Never understood it

PR #4230, opened 2022-12-14
salt PR (lets mycus have regular hair)

LiteralMushroom:
## About The Pull Request
Lets mycus have hair. And beards.
Just say it's mold or something.

Doesn't go the other way around, non-mycus still can't use the caps.

MIGHT change the body sprite. Might.
But not the head sprite, I will allow them to continue being literal dentbrains.

## Changelog
:cl:

PR #4687, opened 2023-07-08
Revert "salt PR (lets mycus have regular hair)"

cdb-is-not-good:
Reverts sojourn-13/sojourn-station#4230

A, Mycus are, per lore- not able to use human and most other sentients to create their body. This PR, if nothing else-  Aesthetically - just makes you a human with a slightly dented head instead of, yknow. An anthropomorphic creature thing with mushrooms instead of a head.

2. Breaks mycus mushroom hairs by giving them a pretty much human head which doesn't fit beneath some mushroom caps

- Why was this even merged? Mushroom people are not waifus. If you want to allow them to take human hair then get a better sprite for their head so that the hair(both human and mushroom) doesn't fucking clip/hang in the air. Was this even tested?


ChefDoggo:
no more mycus waifus, very nice very cool need this now 👍

LiteralMushroom:
The dented head actually predates this PR.

LiteralMushroom:
make this PR if you want (I don't play on Sojourn anymore so there's no real reason for me to care) but. it's not going to fix the thing you're complaining about???

cdb-is-not-good:
> make this PR if you want (I don't play on Sojourn anymore so there's no real reason for me to care) but. it's not going to fix the thing you're complaining about???

You are correct and very handsome for pointing this out. I had read the Pr desc and assumed it was as it claimed responsible for the 'possible change of the body'. I guess I'll need to chase down the actual source of the issue.

PR #4709, opened 2023-07-13
Ports some erismed tweaks

cdb-is-not-good:
https://github.com/discordia-space/CEV-Eris/pull/8216

Hooooboy. Full credit to Humonitarian.

Suffice it to say, I did not port the integration with their mutation system(which was blessedly only one line it looks like.)

changelog quote

Makes Erismed 4 significantly more lethal, generally encourages organs to die more often. Adds a lot of helpful defines.

Changes include:

Reduction in the health of most small organs
Creation of defines for such health values
Change to fracture code, so instead of fractures being required before organ damage, the damage is instead directed to the bones until they break Limb internal organ lists now associate organ with their size for weight-based targeting

Edit; after some testing it appears that trashpile cause some direct(but very minor) organ damage. Instead of paring back the danger of playing in unknown and disgusting trashpiles, i've made gloves 100% protect from scratches, instead of being a 95% chance. Who's idea was that anyway. 

Edit 2: The trashpile organ damage was apparently due to them having sharp + edge. They no longer will cause direct organ damage. Additionally, infections have been tuned back and now give twice as long before spreading to reduce the amount of times you'll die of massive infection in minutes. 

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


cdb-is-not-good:
Worth noting that this should, if anything- be testmerged so it can...yknow. Be tested. I gave it a basic runthrough and nothing seemed immediately broken but yaknow.

cdb-is-not-good:
Given that it's been quite some time and no one has exploded, I think we can merge this for noiw.

PR #4736, opened 2023-08-06
Gives salvagers combat gloves in their locker

DerFlammenwerfer:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>
For some reason, the combat role gets free shock-proof work gloves instead of the role that might actually need them for digging through trash piles. This fixes the problem.

	
<hr>
</details>

## Changelog
:cl:
fix: Salvagers now get combat gloves in their lockers, for safely digging through trash.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/cargo.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/cargo.dm
@@ -344,6 +344,7 @@
 	//armor
 	new /obj/item/clothing/suit/storage/scavengerarmor(src)
 	new /obj/item/clothing/head/helmet/handmade/scavengerhelmet(src)
+	new /obj/item/clothing/gloves/thick/combat(src)
 	//longarms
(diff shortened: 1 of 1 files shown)


cdb-is-not-good:
I might recommend taking the combat gloves away from prospectors to give to Salvagers.

DerFlammenwerfer:
> I might recommend taking the combat gloves away from prospectors to give to Salvagers

And I will refuse your recommendation if you do. Just like the salvaged armor and helmet, these gloves should be considered part of the Prospectors' standard equipment. All other Prospector roles have them, there's no reason for Salvagers to be left out, or to trade one problem for another.

king5327:
I'm gonna peg this as good to merge. Maybe Salvagers shouldn't have combat gloves but is it that big a deal?

cdb-is-not-good:
Its honestly fine, it's whatever. They're just fancy insulated's.

PR #4737, opened 2023-08-06
Revert "Ports"

cdb-is-not-good:
Reverts sojourn-13/sojourn-station#4711

Removes: Stat obscuring
Removes: Extra gym machines that weren't mapped in


Trilbyspaceclone:
I dislike this being reverted, like theirs even a free perk in loadout to not only solve the `i want to see my stats in game` but also has other mechanics, ontop of the gym buff so that anyone with a small price tag can see their stats.

PrettyPsychoForAWolf:
> I dislike this being reverted, like theirs even a free perk in loadout to not only solve the `i want to see my stats in game` but also has other mechanics, ontop of the gym buff so that anyone with a small price tag can see their stats.

At that point it doesn't matter when the players openly mentioned it quite often in the server proper they hated it.

cdb-is-not-good:
I don't consider "It's bad" to be valid criticism. I expect to hear articulate reasons for dislike or it will be disregarded(as it has thus far been). The reason this PR is up primarily is that at least one person has articulated a reason for not liking the changes, even if the reasons were a bit flimsy in my opinion.

SkinkHunt:
It is disliked because its no longer easy to keep up with and be knowledgable with your stats, esspecially since we don't have a nice easy reference guide to establish what will pass or fail checks. In my case, I pay attention to stats to keep an eye on balance, and this change has made it more tedius to do. To expect everyone to perfectly articulate this is an exercise in needless frustration and is frankly unreasonable. 
As discussed in a little impromptu meeting, we should at least have some kind of reason for PRs, as in to fix an issue, rebalance, or augment gameplay. Something, anything. The only reason I've gotten for it, is that it was suggested and ported. Fine. Cool. But if its that easy to merge, its that easy to unmerge, this song and dance really isn't necessary for such a minor change

benj8560:
I'd also like to point out on the topic of ''good reasons to remove it'' many have pointed out there is far from any good reasons to keep it. Regardless I can think of a few. Are stats are highly inflated and for example things like robustness, mechanical and hell all skills matter in the high end for things like timed crafting and ATK usage or hammering airlock bolts.  These named representations cap out at what. 60 or 80? regardless we go up to 150 and it matters in those higher numbers.  Rarely but it matters but I'd like to know what I got rather then guessing.

I also don't see how the entire community whining about something is not a good reason in of itself. Give us a better argument on why its a good thing rather then forcing us to provide a reason it's not. So far all I've gotten is ''The person who made it asked for it to be ported'' As the only real argument for it being here

cdb-is-not-good:
>It is disliked because its no longer easy to keep up with and be knowledgable with your stats, esspecially since we don't have a nice easy reference guide to establish what will pass or fail checks.

Is that really needed, though? If you know that you're not skilled enough for a check then you know your character can't do it. What exactly importance does it have if you failed by one point or one hundred? So you can attempt to bypass it?

>PS: Forcing people to lose out on a perk slot just to see their stats is functionally, a downgrade.

Correct. 

>these named representations cap out at what. 60 or 80?

There is no skillcheck higher than 80. Stats are also softcapped at 80, not providing much if any benefit after.

>regardless we go up to 150 and it matters in those higher numbers. Rarely but it matters but I'd like to know what I got rather then guessing.

No? Just no. Not correct.

>I also don't see how the entire community whining about something is not a good reason in of itself. 

' ^ ' IF the entire community complains but refuses to give reasons, it means either the entire community is unable to express themselves, or they're intentionally not saying why they really want it gone because the argument will look bad or make them look bad. 

>So far all I've gotten is ''The person who made it asked for it to be ported'' As the only real argument for it being here

Generally speaking, we put more weight behind the content made for the server than for the people who demand it to be removed for this reason or that reason. If it can be fixed, tweaked, or made to work - then it should be.

cdb-is-not-good:
I was going to write out a whole snarky thing as to why I'm not really sold on the arguments made here and in OOC, but I already did that repeatedly so here's what we'll do instead. Since you lot couldn't be bothered to really make a point against this PR, we'll do an emojiocracy on it so I can truly wash my hands of dealing with it.

faaaay:
> I dislike this being reverted, like theirs even a free perk in loadout to not only solve the `i want to see my stats in game` but also has other mechanics, ontop of the gym buff so that anyone with a small price tag can see their stats.

The perk is basically a shortcut. Hiding stats just means people who can code-dive get the perk for free.

cdb-is-not-good:
After failing at every single stumbling block to offer reasons for removal the folks who wanted this PR to go through at least managed to win the popularity contest. I am still recommending this PR not be merged, however. 

image

PR #4741, opened 2023-08-14
Fixing 14.5x114mm

Chevalt:
fixes one caliber being too weak compared to others. The 14.5x114 had very much decent armour piercing, but due to various changes around how PVE and PVP worked in the duration between when it was introduced and now, it's fallen a lot behind other weapons that were updated. This weapon is a single shot, meaning it's difficult to use in PVP effectively, making as much damage (Only around 50 brute without the scope) as two regular bullets from most higher end automatic rifles, which just does not cut it for what is supposed to be the best (or one of) possible ballistic precision rifle available. Not only that, but on the PVE side of things, it also was very bad due to the low damage. The AP made it possible to kill most pests easily, but it still lacked against heavier mobs, which normally would be the prime target of a rifle like that. So this PR fixes it with minor changes to the AP and damage values.

### DIFF
--- a/code/modules/projectiles/projectile/bullettypes.dm
+++ b/code/modules/projectiles/projectile/bullettypes.dm
@@ -637,8 +637,8 @@
 
 //// 14.5×114mm Anti-Materiel Rifle Rounds ////
 /obj/item/projectile/bullet/antim
-	damage_types = list(BRUTE = 60)
-	armor_penetration = 100
+	damage_types = list(BRUTE = 80)
+	armor_penetration = 200
 	nocap_structures = TRUE
 	//stun = 5
 	//weaken = 10
@@ -649,7 +649,7 @@
 	recoil = 40
 
 /obj/item/projectile/bullet/antim/lethal
-	damage_types = list(BRUTE = 45)
+	damage_types = list(BRUTE = 60)
 	embed = TRUE
 	armor_penetration = 60
 	agony = 100
@@ -671,7 +671,7 @@
 	recoil = 40
 
 /obj/item/projectile/bullet/antim/scrap
-	damage_types = list(BRUTE = 41.5)
+	damage_types = list(BRUTE = 45)
 	armor_penetration = 50
 	affective_damage_range = 8
 	affective_ap_range = 8


cdb-is-not-good:
Fuck no lol

Chevalt:
we love constructive criticism from CDB

SkinkHunt:
We'll try it out.

cdb-is-not-good:
Okay, here's constructive criticism for you. This is yet another in a long line of PRs buffing ammo instead of addressing bloated values elsewhere. You want to powercreep this bullet up in damage instead of addressing why it's not good enough(which, let's be clear; is not true. It's fine in its particular use case, as a high-armor cracking rifle). Armor values too high? Other bullets too good compared to it? That doesn't mean this needs a buff, it means those need to be dialed back.

Chevalt:
> 

It's not powercreep, it's genuinely underpowered. It's not just an "armour cracking rifle", it's an anti-material rifle, it's meant to destroy big things, like heavy machinery for example. The measly 50 damage is only equal to two .50 kurtz rounds, which can be found practically anywhere, and aren't even that powerful in many cases, which just doesn't fit this rifle that if it made sense wouldn't need as many shots as any regular shotgun with slugs to kill someone. If we're talking balancing, are you gonna change this one niche rifle so that it fits with all the others, or are you gonna nerf for it EVERY other rifles by "dialing them back" as you say? Not only that, but many other rifles have comparable or semi-comparable damage and armour piercing values while being much more common and having a higher magazine capacity.

cdb-is-not-good:
If it fails to fill its niche, then instead of just pumping the damage(which doesn't actually fix the issue you're complaining of.) find a way to make it do a better job as a 'big boy stomper'. Giving it 20 extra damage WONT help that, it WILL make it all 'round better. We have super-effective damage modifiers, consider finding a way to make it that this weapon will do extra damage to heavily armored things.

Chevalt:
> If it fails to fill its niche, then instead of just pumping the damage(which doesn't actually fix the issue you're complaining of.) find a way to make it do a better job as a 'big boy stomper'. Giving it 20 extra damage WONT help that, it WILL make it all 'round better. We have super-effective damage modifiers, consider finding a way to make it that this weapon will do extra damage to heavily armored things.

The extra pen does that actually on mobs.

cdb-is-not-good:
I disagree whole heartedly with buffing an already powerful murder gun. Find a better way to make it perform in its niche if you think that's what it needs, but do not just give it an across the board damage buff.

PR #4749, opened 2023-08-22
Fixing the previous mapping PR. (spaceport)

Chevalt:
Fixes the underplating being present below two sets of stairs, also fixes there being no plating in some areas of the spaceport. Adds a few more trees in some areas because it'd look cool.


cdb-is-not-good:
I will assume this is working and trust you that it is not fucked up

PR #4754, opened 2023-09-05
Bullet balance

cdb-is-not-good:
Tweaks a few things about bullet balanz.

HP round have had their post-pen multipliers increased slightly, basically - while 9mm will still only do about double damage against an unarmored target, a 12mm will do 2.5x instead.

HP rounds have also had their recoil adjusted a bit, they now generally have the same recoil as a standard round.

Pistol effective ranges increased. 9mm up to 5(2 tiles away from screen edge)from 3, 10mm 6(also from 3), 12mm 7(also from 3). their HV variants have 1 extra tile of effective range. Who's idea was it even to say that pistol calibers would have THREE TILES of range before getting damage falloff??

despite initially wanting to make buckshot more effective at a less stupid range, i've instead given it 1.5x postpen mult to bring it in line with HP rounds, this will VERY likely need tweaking.

Slugs are now effective up to 7 tiles because why the FUCK WOULDNT THEY BE

Biomass slugs as a child of regular slugs have the same.

Scattershots also had 2x postpen applied, really will just TEAR through unarmored targets like poor civilians.

8.6 has had its recoil raised slightly across the board

Baroque is now not stupid. AP down from 250 to 50. Supereffective mult of 6, total damage is roughly the same if not a little lower but ONLY against /mob/living(other than humans.). 

Also no longer embeds because why would it? It's a high power hitscan that pierces walls, it goes RIGHT through you. 

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullettypes.dm
+++ b/code/modules/projectiles/projectile/bullettypes.dm
@@ -693,7 +693,7 @@
 	nocap_structures = TRUE
 	damage_types = list(BRUTE = 40) //Grab me musket as the founding fathers intended
 	armor_penetration = 50 //no longer a little jank, much like other older rifles it falters in terms of AP while still having enough to really smash through armor.
-	supereffective_mult = 5 //we do 40 damage base, up to 200 with supereffective.
+	supereffective_mult = 6 //we do 40 damage base, up to 240 with supereffective - plus AP bonus, plus agony bonus, about the same 350~ as before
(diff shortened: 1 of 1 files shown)


cdb-is-not-good:
Im way too tired to deal with this rn, but I'll also likely bundle in some changes to the bone rifle bullets to make them not NEED 100 billion AP to be effective against mobs. I don't really believe the AMR needs a buff of any kind so I don't intend to bundle that in unless someone can give me a convincing argument that it needs tweaking

PR #4755, opened 2023-09-05
Events 'n Objectives.

### LINKED ISSUE #4771: Core implant objectives.
so appears you can get the convert objective with a psionic organ and I assume any core implant. 
dreamseeker_BuAmbAaugM

Not bug related but this objective should honestly just be removed or have it's limited antag status removed? Force conversions really don't fit are church lore.

cdb-is-not-good:
Seems to have FINALLY fixed* infestation events causing mobs to go to burned outpost, spider outpost, etc.  Due to the way this fix works(by marking these areas as "maintenance" this may cause some unforeseen consequences, I have tested it about as well as I can but that's worth noting. Blame this system

Maybe fixes the issue with hiveminds/blobs spawning in dungeons like preppers. Seems to have at least limited both to spawning on colony zees instead of sometimes popping up in preppers 'n such. They do sometimes spawn outside, still but that's ....okay, i guess. 

Additionally, many factions had all their individual objectives returned - including the semi antag ones. Now you'll be bribed with tasty, tasty level ups in exchange for being a bastard. It goes without saying that Marshals did not get given the 'be a shithead' objectives back for reasons that should be obvious(massive access, authority, etc), if you lot want them enabled for Marshals then we can I guess consider it?

Some of these will probably have bugs and some of them may need to be tweaked or considered for removal, but overall I think they should offer some incentive for players to be a lil naughty in exchange for level ups - or not!

fixes #4771

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/controllers/subsystems/migration.dm
+++ b/code/controllers/subsystems/migration.dm
@@ -211,9 +211,9 @@ This proc will attempt to create a burrow against a wall, within view of the tar
 
 	switch (GLOB.storyteller.config_tag)
 		if ("jester") // Jester will most likely not reroll the maintenance area check.
-			reroll_prob = 19.5
+			reroll_prob = 99.5 //loljk, original values caused issues with infestation event.
 		if ("warrior")
-			reroll_prob = 80
+			reroll_prob = 99.5
 
 	//Lets copy the list into a candidates buffer
 	var/list/candidates = GLOB.all_burrows.Copy(1,0)
--- a/code/datums/objective/individual_objective/church.dm
+++ b/code/datums/objective/individual_objective/church.dm
@@ -27,7 +27,7 @@
 	if(completed) return
 	UnregisterSignal(mind_holder, SWORD_OF_TRUTH_OF_DESTRUCTION)
 	..()
-
+/*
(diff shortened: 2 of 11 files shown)


cdb-is-not-good:
A small note is that this, if it actually works this time, will be a notable difficulty adjustment as these things do in fact happen rather frequently especially on game modes like warrior. Hopefully the bullet buffs(and the time players have to escape) should be adequate enough to deal

du-top:
The outposts were seen as normal areas? Possible that they counted against the "dirty areas" points shown at round end. Which are completely insignificant, but always a bit insulting.

cdb-is-not-good:
That may very well be the case actually yeah. Yes, they didn't have any special markings other than the one that designates whether or not they can be used for the prospectors strip objective

king5327:
If this works, then it should be good to merge. But I can't see enough of it from the code to put the tag on it.

cdb-is-not-good:
After testing it seems to finally work. At least so far, it's done as expected and has had 0 complaints.

PR #4759, opened 2023-09-08
PROBABLY fixes certain deferred boxes being an obnoxious meme.

cdb-is-not-good:
storage set their own inventory size/max itm size based on what's in them, cool right? Well, this means things like the tool-mods kit from lonestar is the most potent storage item ever because it comes with 12 normal size slots. They can now only hold the kinda items they come with, I'm very likely missing some stuff but oh well.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


king5327:
Why not disable the automatic scaling for those containers?

cdb-is-not-good:
> Why not disable the automatic scaling for those containers?

Probably an equally good solution, I had thought partly to just say you can never ever put stuff back in them tbh.

PR #4764, opened 2023-09-10
Enable spoons, forks and knives in lunchboxes

du-top:
## About The Pull Request
<details>
<summary>
	Enable spoons, forks and knives in lunchboxes
</summary>
<hr>
Spoons, forks and knives should fit in a lunchbox, so you don't have to eat your steak like an animal. As we don't have dinner knives / menu knives and _every_ knife is based on the kitchen knife, this means that every knife works. Other kitchen tools are not allowed.

image

	
<hr>
</details>

## Changelog
:cl: dutop
add: Allow spoons, forks and knives in lunchboxes.

/:cl:


cdb-is-not-good:
Based.

PR #4779, opened 2023-10-04
Creating an ERT Ready Room

Pawoverlord:
## About The Pull Request
<details>
This creates and adds an ERT Ready Room.
<summary>
	This creates and adds an ERT Ready Room that will allow admins and mods to access and prepare ERT characters quicker and without needing the ADMIN+ or SPAWN+ perms. This is exclusively for the times that we do need and use ERT when applicable. 
</summary>
<hr>

The room, seen below comes with the capacity of five units. Three Marshal Units and two Corpsman Units. Images of the loadout are provided with this. For ammo, they are given an ammo printer with six stacks of refined scrap which should be more than enough scrap for any of the loadouts. They come with WO headsets, Marshal IDs, and ammo pouches for each type of weapon. 

There's four boxes of gun parts for all the attachments for any of the guns, ten medium super cells, ten small super cells for the respective guns. Three medium and three small atom cells. 

image

image

image

The gun choices for the ERT team is designed with the idea of low-tier threats towards high-tier threats, left to right accordingly. Grenades are also provided as a tactical resource. Though, I excluded adding anything like RPGs as I felt that ERT should not need to resort to RPGS, as they would just be Death Squad. I additionally included a sectech for any security items such as handcuffs, maglights, or flashes. 

image

image

image

image

	
<hr>
</details>


Pawoverlord:
Err I think I goofed and my commit didn't bring the mapfile over. Heehoo I'm dumbo.

PR #4785, opened 2023-10-06
Add linen bin to Janitor closet

du-top:
## About The Pull Request
<details>
<summary>
	Adds a linen bin (bedsheetbin in code) to the Janitor closet. Enables the linen bin to contain double-sized bedsheets
</summary>
<hr>
I simply wanted to add a linen bin to the Janitor closet, to be able to change the bed sheets in Dorms. Collecting all sheets, washing them and returning them is just not great, as it leaves the beds uncovered! Having no spares sucks, as well, because maybe someone needed a bedsheet to cover up a shivering friend somewhere.
Problem: The linen bin does only provide single bed sheets, while the beds in the dorms are double-sized. So I simply added a var "double" that defines the bin as a bin for double-sized bedsheets. Default is FALSE. Var can be set in map. Description of linen bin shows if it's loaded with double sheets or not. Tested on local server.
Note: You COULD click on a double linen bin with a single-size bedsheet and load it into it. Examining would say there are 21 double bedsheets. Don't do that. You'd look stupid.

image

	
<hr>
</details>

## Changelog
:cl: dutop
add: Double-sized linen bin to Janitor closet.

/:cl:

### DIFF
--- a/code/game/objects/structures/bedsheet_bin.dm
+++ b/code/game/objects/structures/bedsheet_bin.dm
@@ -283,18 +283,28 @@ LINEN BINS
 	var/amount = 20
 	var/list/sheets = list()
 	var/obj/item/hidden
+	var/double = FALSE
 
 
 /obj/structure/bedsheetbin/examine(mob/user)
 	..(user)
 
 	if(amount < 1)
-		to_chat(user, "There are no bed sheets in the bin.")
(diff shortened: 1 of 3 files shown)


cdb-is-not-good:
yolo, hope it works

PR #4813, opened 2023-10-20
Organ vender spawn rate buff.

benj8560:
## About The Pull Request
After many a long days spent as medical in deep maints looking for OMG venders I've become TIRED. 

Significantly improves the spawn rates of all the useful organ venders. About 2 times more common

Reduces the spawn rates of the classics one since it just exists to sadden me every time I see it. (same stuff medical can produce round start) Reduced by a LOT! Hate you classics vender.

Reduces the spawn rates of medi sarcophagus sleepers. Still very common! Lots of deep maint maps contain these hardspawned. Reduced by about a quarter

Mostly the logic behind this is these numbers no doubt come from Eris where deep maints is a walk in the park compared to here. This should be much more inline with the level of difficulty to reward down there. Plus hopefully! Gives are bored doctors more to do.

Ideally test merge to see this in practice I could not seem to get deep maints to generate on a private sever and could not be bothered to figure out why. Only testing I did was by manually placing spawners.

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/modules/aberrants/organs/design_disks/disks.dm
+++ b/code/modules/aberrants/organs/design_disks/disks.dm
@@ -40,7 +40,7 @@
 
 /obj/item/computer_hardware/hard_drive/portable/design/omg/teratoma
 	disk_name = "Oh My Guts! Bespoke Teratomas"
-	license = 10
+	license = 20
 	designs = list(
 		/datum/design/organ/teratoma/special/chemical_effect,
 		/datum/design/organ/teratoma/special/stat_boost,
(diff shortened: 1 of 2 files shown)


benj8560:
small change adds 10 extra uses to the teratoma disks now totaling 20 (found in hacked deep maint venders) Leaves the prefabbed organ disks alone however they seem to be pretty good for the effort involved. I'd mostly just like to see players be a little less cheap with these printing lots of them to explore just what things can be done. Rather then printing the few they know are worth using. These numbers are likely once again originally balanced around Eris where disk cloners are easy to come by.

PR #4816, opened 2023-10-23
Updated fortress

CactusMouth:
## About The Pull Request

Added more loot, mainly. These include meds, armour, weapons and similar.

<hr>

Currently mobs are unchanged, though can be reduced if requested. Some smaller issues such as walls and mines are fixed, along with the added gear. More loot can be added on demand, but currently it is unlikely to be added for a few days.


cdb-is-not-good:
p sure this is a copy of another PR that was already merged. In the future, if you make a new PR because of conflicts close the old one.

PR #4818, opened 2023-10-24
Update - Loot & Mines

CactusMouth:
You know, when I took up mapping I did know it would be miserable. However, the word 'miserable' has it's limits. Also I am sorry Cdb but making an entirely new PR is the easiest way for me to get around the multi-file problem quickly

## About The Pull Request
] Replaced Excel mines with normal mines due to runtime issues (please get back to me on that one)

] Added more loot

I may add more bears and excel slaves in the future, depending on if this is enough to satisfy the masses. Clearly you are all too cowardly to break into somewhere without enough reward.

Enjoy.


CactusMouth:
Fuck

cdb-is-not-good:
Speedmerging this since it's a bugfix for a current PR

PR #4821, opened 2023-10-26
Greyson Offices changes

Rat-Time:
Attempting to make the rooftops of Greyson Field Offices map interesting rather than a 5 minute detour.
Loot has been moved, not added.
Additional methods of getting around the map have been added and are guarded accordingly.

A fancy landing pad, walkways, catwalks, poor structural integrity, and Greyson AI generated bathrooms ahoy.


Trilbyspaceclone:
image
no door?
also dont have lattis on grass/below level

Trilbyspaceclone:
image
also
the joke of the maze is that is the bathroom/breakroom/printer room

Also random locker in the wall
Also also dont have walls that are going like that in the circuals without full 3 tiles so they connect properly

Rat-Time:
The lattice is meant to indicate skybridges above but if it causes some kind of problem then they can be taken out

Rat-Time:
Got rid of that locker, added the walls, and changed the bathrooms to storage.  I checked the phantom door and it seems to work.  I just tested it and it showed up where I put it, was hidden under that invisible light maybe?

Rat-Time:
Upped the difficulty slightly, removed a duplicate oddity gun that was hiding, reduced the amount of guaranteed lockers.

PR #4826, opened 2023-10-28
Joseph atton found the ids!

Dongels:
## Changelog /  Map change
:cl:
Added a box of Spare ids in the premiers room. Seems the last premier forgot to put it back up on their table when it was knocked off during their last drink. Goofy thing was **under the table**. 

*SMH*

:cl:


cdb-is-not-good:
👍

PR #4829, opened 2023-10-29
Soil crafting crafting time reduction

WilsonWeave:
Quality of life change. Makes the entire process of making soil plots less mind numbingly slow.

It just tweaks one number. It compiles.


cdb-is-not-good:
I'd prefer the change be to 45 instead of 20.

cdb-is-not-good:
yolo. I'll nerf this if it becomes a thing

PR #4831, opened 2023-10-30
Adds risky CHECK_TICKs to human rendering code

nikothedude:
## About The Pull Request
This will definitely make char creation slower, but will definitely make the server faster. IT might also break things horribly.

	
<hr>
</details>

## Changelog
:cl:
code: Human rendering is now heavily CHECK_TICKed
/:cl:


nikothedude:
for the record CHECK_TICK is a macro that sleeps the proc if the server is over its tick limit, this prevents lag spikes at the cost of delaying proc execution

benj8560:
Seen a few strange happenings with this sadly no screenshots but in the span of 20 minutes or so I seen what looked like corrupted images for the character example in the setup. Weird lines, chunks of the image missing/moved to the side weird janky lines etc etc. However I only seen this the once 

Also weirdly had airlock icons showing up in there very random and strange...but went away soon after.

Pretty minor stuff I'll try and grab a screenshot if it happens again but overall did not prevent me from doing stuff so no biggy.

benj8560:
jvfkFuQYf9


Screenshot obtained

PR #4838, opened 2023-11-04
Add projectile tag to rpg

Chevalt:
Does what it says on the tin, this adds a projectile firing tag to the RPG allowing it to be modded like other projectile guns.

### DIFF
--- a/code/modules/projectiles/guns/projectile/other/RPG.dm
+++ b/code/modules/projectiles/guns/projectile/other/RPG.dm
@@ -21,7 +21,7 @@
 	init_recoil = HANDGUN_RECOIL(3)
 	bulletinsert_sound = 'sound/weapons/guns/interact/china_lake_reload.ogg'
 	twohanded = TRUE
-	gun_tags = list(GUN_SCOPE)
+	gun_tags = list(GUN_SCOPE, GUN_PROJECTILE)
 	allow_racking = FALSE
 	auto_rack = TRUE //so if we get loaded were good
 	serial_type = "SA"


cdb-is-not-good:
😨

Won't this allow for silliness like putting a gauss on the RPG?

cdb-is-not-good:
actually, more import and real question. Why? Even if we ignore potentially busted additions won't most mods on the thing be....worthless? It gets 1 shot, it's DEFINITELY not going to be allowed to have an overshooter, and each shot has FAT recoil IIRC?

Chevalt:
> actually, more import and real question. Why? Even if we ignore potentially busted additions won't most mods on the thing be....worthless? It gets 1 shot, it's DEFINITELY not going to be allowed to have an overshooter, and each shot has FAT recoil IIRC?

overshooter uses internal mag so that won't fit

benj8560:
personally against this. You don't need to mod a damn RPG up your already pretty much guaranteeing you win the fight if you land your shot as is. It's also wildly strange for modifications intended for small to medium caliber weapons to fit on one.

cdb-is-not-good:
> > actually, more import and real question. Why? Even if we ignore potentially busted additions won't most mods on the thing be....worthless? It gets 1 shot, it's DEFINITELY not going to be allowed to have an overshooter, and each shot has FAT recoil IIRC?
> 
> overshooter uses internal mag so that won't fit

Still waiting to hear what the point is of this change. Why is allowing an RPG to take mods a desired outcome? what benefit does this have?

cdb-is-not-good:
If I don't get a decent answer to the question i'll be denying this PR.

PR #4840, opened 2023-11-04
Updated fortress - SPACE BEGONE

CactusMouth:
Replaces the space below the lower walls in the Excel Fortress with flooring. That's it, stay safe.


cdb-is-not-good:
still got merge conflicts.

PR #4841, opened 2023-11-05
Nanopaste tweaks.

cdb-is-not-good:
Nanopaste has received the balance stick. Despite telling everyone repeatedly that nanopaste didn't work it turns out I am a liar and stupid too(it works, it's just that people frequently come into medical with conditions that prevent their revival other than the timer). I'd RATHER kill this in its crib, but I cannot.

Nanopaste now requires the surgical perk to use(it's LITERALLY brain surgery) and additionally uses all 5 charges instead of just 1/5 when fixing someone. Additionally, upon being nanopaste'd you receive an even more extreme version of rez sickness that won't go away on its own - your reminder for the remainder of the round to take proper precaution next time(If you try to bypass this by going down and coming back up, I'll be very cross.)

Also tweaks up the nanopaste code a bit. There's warnings if the nanopaste lacks enough nanites to proceed or if you haven't got the right training for it.

Oh also makes the discomfort temp a bit lower so that fevers will trigger it.

Drafted for now - A few features aren't working and the discomfort temp stuff needs testing.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/__DEFINES/perks.dm
+++ b/code/__DEFINES/perks.dm
@@ -101,6 +101,8 @@
 #define PERK_REZ_SICKNESS /datum/perk/rezsickness
 #define PERK_REZ_SICKNESS_SEVERE /datum/perk/rezsickness/severe
 #define PERK_REZ_SICKNESS_FATAL /datum/perk/rezsickness/severe/fatal
+
+#define PERK_FSYNDROME /datum/perk/fsyndrome
 //////////////
 //Race Perks//
 //////////////
--- a/code/datums/perks/job.dm
+++ b/code/datums/perks/job.dm
@@ -342,6 +342,20 @@
 	if(holder.buckled)
 		cooldown_time -= 2 SECONDS
 
+/datum/perk/fsyndrome
+	name = "Flatline Syndrome"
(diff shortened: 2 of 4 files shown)


SkinkHunt:
"Its literally brain surgery" 
Well, I mean, according to the description of when you use it, you're literally just putting them in and the nanites do the rest. It doesn't matter where in the body you put them, it just has to be through an open incision as well. 
 At least this is how I understand it. If you are going to go with this change, I would suggest changing the use description as well and adjusting the use code so that it only fixes brain stuff when used on an incision in the head.

cdb-is-not-good:
It still works for those other uses without any kinda check, but neverminding the brain surgery bit it *is* an incredibly potent thing. That said, I also didn't make it any rarer- and as "rare and expensive".....not really? 3 exist in the medivends for 300$ each which is pretty minor

SkinkHunt:
For some reason I thought they were currently 800, lmao.

cdb-is-not-good:
Tiny amount of difficulty making PR work as i want, scrapping this for now.

PR #4845, opened 2023-11-06
Fixing the commas

Luxyleathehat:
Compiled this time around with the added commas that my tired ass brain missed


cdb-is-not-good:
already fixed..........................................................

PR #4847, opened 2023-11-07
Rig Module Damage Change

Dongels:
:cl: Attempt to fix the flimsy module issue holding back people from using the Rig suits out of fear of loosing modules too fast. 

(Look over code & Test merge first)
_Honestly not sure if this is how it should be changed code wise, but I did test it..and seems fine on my end._


cdb-is-not-good:
Discussions indicated that testing showed this changes modules to be unbreakable. Please either remedy this or continue testing to find the limit of damage that causes a module break.

cdb-is-not-good:
Still doesn't work, no changes made. I've included a change in one of my PRs that should raise "health" of rig modules.

PR #4859, opened 2023-11-15
Eris plasma port and balance

benj8560:
A close to direct port of eris's current state of  plasma weapons. Including their (really badass) sound effects also accounting for psionic variants of them.

Purger and dom now have very significant differences to make them much more distinct from one another

Dominion now exiles at dealing with large amounts of unarmored targets with the exception being its AOE fire setting. Its standard settings although extremely spamable have 0 AP. AOE setting has remained exactly as we had it. Apparently.

Purger does better with armor having a low fire rate full auto setting and a very very inefficient AOE setting much weaker then the dom.

Both weapons now have recoil that needs considering when modding especially the purger.

Cassad is now akin to a bigger Centurio. Its ion setting being less energy efficient however

Unseen Sun as expected is slightly worse in every regard to a real Cassad (still pretty nice tho) with the deep maints version being a exact copy.

All of these weapons above have been blocked from using unmakers. They are inherently DPS monsters without. 

Got mildly worried about rendering the Carpediem and Antebellum totally pointless. Both have had their gold cost reduced down to 1 sheet (0.5 in a bioprinter) to hopefully still retain their value as a cheap mass printable gun. Gold is a lot easier to come by then uranium after all. Stats otherwise left alone

Concillium has sat as being a absolutely useless gun when looking at the balance of the server over all. I personally think it needs more then this but for now only reducing the plasteel cost down to 20 opted to not touch its charge use as it pretty much makes it a sunrise but better if I do. Now has slightly high AP and damage as well as reduced recoil. Hopefully the heavy hitting las MG it's suppose to be.

As some attempt at fairness guild is seeing some buffs. 

Reductor has had its thunder shunt mod block removed. The block was added before are massive nerf to fire rates and hopefully should help even this mod out. Keeping its unmaker block tho. It's also seen a fair buff to firerate nothing to crazy I think but I noticed it was one of its major drawbacks. 

The code for it's overheat mechanic has been commented out now honestly? I like the overheat mechanic for balancing the problem is it's far to unforgiving on the reductor and I don't exactly know how to tweak it. Even if it WAS more forgiving it does not even work as far as I'm aware you can fire all day long without any issues. Commented it out in hopes someone might fix it some day.

The Slab has had it's Superkill setting swapped for a Supercharge setting that hopefully works fine something most agreed would actually make the gun interesting and worth using. Pretty much a charge setting like on the spider rose. Seems to work fine! Fully charged putting out a nice even 50 damage. Still not outstanding but at least used correctly you could kill something big.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


benj8560:
riiight I personally feel I'm done here but I'm gonna let this cook for a day while I sleep see what those who care think.

cdb-is-not-good:
a bit iffy on this since it looks like a buff Im not 100% sure is necessary, but unless another reviewer has something to add we'll testmerge and give it a go.

benj8560:
Pretty much everyone agrees the church plasma weapons are just underwelming and overpriced. They CAN be effective modded just right as is but usually just a way to flex you have alotta mats. It is a buff...mostly. I'm not sure exactly how debilitating these lower AP or no AP plasma shots will be but i imagine these will struggle with some megafauna and be more ideal for large amounts of small bug killing and PVP

benj8560:
Centurio is a comical example of this. 2 plasma at T4 for a absolutely crazy ass plasma pistol. These numbers pretty much compare to it while being a touch better due to being far more expensive and rifle sized

cdb-is-not-good:
> Centurio is a comical example of this. 2 plasma at T4 for a absolutely crazy ass plasma pistol. These numbers pretty much compare to it while being a touch better due to being far more expensive and rifle sized

counter argument, centurio requires research + materials and every point spent on the centurio is a point not spent on 'POSIS POSIS POSIS, RESEARCH POSIS? STOCK PARTS?'

benj8560:
> 
> counter argument, centurio requires research + materials and every point spent on the centurio is a point not spent on 'POSIS POSIS POSIS, RESEARCH POSIS? STOCK PARTS?'

Those materials being steel and a tiny bit of plasma. Research cost is pretty much irrelevant currently since everyones started using the ADMS finally. I usually b-line for this gun since it's just a absolutely crazy all round use weapon. A single hatton tube recycled from the trash gets you like 5 of these guns or something.

benj8560:
plus fits in a holster and can be duel wielded and! Almost no recoil. Honestly it still outclasses these rifles when duel wielded

PR #4860, opened 2023-11-16
Adds the disk storage fridge

SingingSpock:
## About The Pull Request
<details>
<summary>
	 Adds a disk storage machine
</summary>
<hr>

Adds a subtype of smartfridge that holds disks, and maps them in at the places where a table/rack that held a faction's disks used to be, as well as the public lathe. More can be made with a board found on the Guild Circuits disk. Also adds the ability to make regular smartfridges for good measure, since those couldn't be made previously.
	
image

<hr>
</details>

## Changelog
:cl:
add: New disk storage fridges
/:cl:

### DIFF
--- a/code/datums/autolathe/circuits.dm
+++ b/code/datums/autolathe/circuits.dm
@@ -108,6 +108,14 @@
 	name = "trade sending beacon"
 	build_path = /obj/item/circuitboard/trade_beacon/sending
 
+/datum/design/autolathe/circuit/smartfridge
+	name = "smartfridge"
+	build_path = /obj/item/circuitboard/smartfridge
+
+/datum/design/autolathe/circuit/smartfridge/disk
+	name = "disk storage"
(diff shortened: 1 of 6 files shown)


benj8560:
minor stuffs but the table next to the prospectors lathe has been rendered inaccessible.
dreamseeker_9Kw264TrDY

 Theres also a recharger causually lodged into the guilds disk storage.
dreamseeker_luWaJh0RDx

benj8560:
on another note. I believe the server was called Oculus a smaller Eris based server that had this exact same idea. They should have sprites that work for these I recall them looking akin floppy disk organizer pretty much exactly what they need to be. Made to be on table.

SingingSpock:
Swapped the positions of the disk fridge and the table in prospies, moved the charger over to the table in Guild, and added a disk storage to the Steward office since I missed they get their own lathe. As far as sprites from another server, I have very little experience grabbing/integrating such but someone else could definitely do it after merge

PR #4867, opened 2023-11-26
finalized disk storage

benj8560:
Some final touches on the prior PR #4860. Adds a more fitting sprite ported from Occulus for the disk storage units one that is more table friendly. 
Also readds the tables under said disk storage units and makes it so you can walk over said disk storage units. 
Gives guild back their second recharger they lost. Trust me they need two. 
Adds a few missing spots for these including the lodge. 
Shifts things around in the R&D lab as well to make the lathe setup a little less silly. 

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


benj8560:
As a side note I'd really like to fix the issue with these being not being unwrenchable or being able to take them apart but that's mostly a issue stemming from normal smart fridges and I have no idea how to correct it but if you DO know let me know!

benj8560:
the icon is content as well :P

PR #4873, opened 2023-11-29
Ports KARL from eris

Guidesu:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>https://github.com/discordia-space/CEV-Eris/pull/8339
Ports KARL from eris, a mining weapon and tool. Already tested in game and it works correctly with a few adjustuments for our codebase

	
<hr>
</details>

## Changelog
:cl:

/:cl:


cdb-is-not-good:
You left out how this works differently. Seeing as the original works more akin to /tg/s miner gun(weaker in atmosphere) what's it like here? Is it weaker? Is it just as strong in all circumstances?

Guidesu:
linters going funky it compiled on my PC

benj8560:
sounded neat so I poked this...seems...cool but useless? The ranged setting is pretty much useless in mining since it's got pressure tho does give fire stacks since it's plamsa. Also seems to lack armor pen for the melee mode and looking at Eris it is suppose to have armor pen. Giving it a bonus damage to golems rather then making it function at full damage in pressure might be the best bet.

benj8560:
degradation rate will also likely need to be reduced by a lot to account for are _fun_ degradation system

Noobetski:
perhaps change it to space salvaging tool? You've got low pressure there and it's pretty good source of mats too

PR #4886, opened 2023-12-13
Sensible greyson suit

benj8560:
Makes the frankly MEH Greyson softsuit actually compact like the flavor text suggests by making it functionally a voidsuit rather then a old style suit. Migrates the code for these over to merc suits as we seem to be using it for misc suits rather then leaving them along side old style suits.  Including the greyson hardsuit.

Gives the suit 100 rad protection as the flavor text suggests it's a ''engineering suit'' why 30 rad weh

Makes sure to account for these changes in other places like venders, loot piles, etc, Midly reducing the spawn rate as you only need 1 roll for the set now. technically makes the Greyson hardsuit slightly more common as a side effect. It's already at 1 so not much I can do about it but a change I personally felt needed to happen anyways after spending entire rounds looking for one.

Greyson Star Traders suit tweaked as well getting the same size bonus as well as 50 rad up from 15. Removes it helm from the vender and adds the helmets price to the suit since its all one in one.

Few small typos I seen along the way fixed as well.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


cdb-is-not-good:
....small. The size of a magazine.

cdb-is-not-good:
normal is already compact compared to a regular suit

benj8560:
right no so....the helmet is not attached to the suit. Helmet plus suit makes it total out to the same size as a standard voidsuit.

benj8560:
kinda a lazy fix better would have been making it have a toggle helmet like Eris has it I suppose I should

benj8560:
changes made to be less cursed. Now is voidsuit like it always shoulda been

benj8560:
and yes I tested it

code/modules/clothing/spacesuits/syndi.dm
```diff
@@ -38,6 +38,7 @@
 	icon_state = "os_spacehelm"
 	item_state = "os_spacehelm"
 	desc = "A helmet produced by GP. It's outdated but sleek design makes it still viable to use even after all these years."
+	w_class = ITEM_SIZE_SMALL
```
cdb-is-not-good, line 41: nix this one too

benj8560: woop

cdb-is-not-good:
mkay that's- fine now.

PR #4887, opened 2023-12-13
Recycling vendor: consider stacks of materials in payout and storage

du-top:
## About The Pull Request
<details>
<summary>
	When using the recycling vendor, consider stacks of materials.
</summary>
<hr>
 When you are using the recycling vendor and add a stack of materials, for example a stack of five plasma cubes, you will get the price of _one_ cube and _one_ cube will be stored into the vendor's storage. To circumvent that, you have to pop one item from the stack, put it in the vendor, click recycle, pop the next one, ... That sucks.

Now the code checks for material stacks. If the object hat has been shoved into the vendor is a stack, the amount is considered. The value is multiplied, the amount added to the internal storage as well.

Here, have a before/after of adding a stack of five plasma crystals. Everyone loves before/afters.

image

	
<hr>
</details>

## Changelog
:cl: dutop
fix: Consider stacks in recycling vendors

/:cl:

### DIFF
--- a/code/game/machinery/vendor/recycle_vendor.dm
+++ b/code/game/machinery/vendor/recycle_vendor.dm
@@ -162,9 +162,18 @@
 		if(i in materials_supported)
 			if(i in materials_allowed)
 				var/material/M = get_material_by_name(i)
-				var/obj/item/stack/material/S = new M.stack_type(null, stored_item_materials[i])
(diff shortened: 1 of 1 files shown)


du-top:
Of course I did only test my addition and didn't notice that I fucked up the recycling of all items that are NOT material stacks. My bad, I am on it!

PR #4918, opened 2024-01-10
Elite Mercenary moved from Simple Mob

CactusMouth:
So for my events, I've been using the Elite Mercenary group of mobs, which are simple mobs. As such, I am moving them to Superior Mobs with some added benefits.

> Mobs are generally strong, so aren't really spam-able for events
> Mobs are also dangerous, with automatic shotguns, revolvers, SMGs and melee. 
> Should hopefully be much more useful in events.

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/human/subtype/solelite/solelite
+++ b/code/modules/mob/living/carbon/superior_animal/human/subtype/solelite/solelite
@@ -1,6 +1,6 @@
 /mob/living/carbon/superior_animal/human/solelite
-	name = "Solfed Elite"
-	desc = "Black and charred hardsuit, sub-machine gun in-hand, a being of absolute terror to traitors of the Federation."
+	name = "Elite Operative"
+	desc = "Black and charred hardsuit, sub-machine gun in-hand, a being of absolute terror to those who dare approach."
 	icon = 'icons/mob/mobs-humanoid.dmi'
 	icon_state = "syndicate_stormtrooper_smg"
 	stop_automated_movement_when_pulled = 1
@@ -49,7 +49,6 @@
 
 	inherent_mutations = list(MUTATION_HEART, MUTATION_LUNG, MUTATION_LIVER, MUTATION_BLOOD_VESSEL, MUTATION_MUSCLES, MUTATION_NERVES)
 
-	drop_items = list(/obj/item/gun/projectile/)
 	faction = "elitemerc"
 
 /mob/living/carbon/superior_animal/human/solelite/handle_breath(datum/gas_mixture/breath) //we have are own air supplies
@@ -64,7 +63,7 @@
 
 /mob/living/carbon/superior_animal/human/solelite/melee
 	icon_state = "syndicate_stormtrooper_melee"
-	desc = "Black and charred hardsuit, sword and shield to match, a being of absolute terror to traitors of the Federation."
+	desc = "Black and charred hardsuit, sword and shield to match, a being of absolute terror to those who dare approach."
 	armor = list(melee = 65, bullet = 65, energy = 80, bomb = 80, bio = 90, rad = 25) //Melee shield
 	ranged = 0
 	melee_damage_lower = 40
@@ -72,7 +71,7 @@
 
 
 /mob/living/carbon/superior_animal/human/solelite/pistol
-	desc = "Black and charred hardsuit, revolver in-hand, a being of absolute terror to traitors of the Federation."
+	desc = "Black and charred hardsuit, revolver in-hand, a being of absolute terror to those who dare approach"
 	icon_state = "syndicate_stormtrooper_pistol"
 	projectiletype = /obj/item/projectile/bullet/rifle_75/hv
 	drop_items = list(/obj/item/gun/projectile/revolver/hornet)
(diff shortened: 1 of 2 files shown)


cdb-is-not-good:
Are these vatgrowns or meant to be standins for actual humans, because if the latter I'd strongly suggest against using them in the future. If the former, I'm fairly certain the Federation didn't ...use vatgrowns in this way, not that i blame you for making use of what you had available.

code/modules/mob/living/carbon/superior_animal/human/subtype/solelite/solelite
```diff
@@ -0,0 +1,100 @@
...
+	return
+
+/mob/living/carbon/superior_animal/human/solelite/handle_environment(var/datum/gas_mixture/environment) //are armor legit is a hardsuit
+	return
+
+/mob/living/carbon/superior_animal/human/solelite/start_pulling(var/atom/movable/AM)
+	to_chat(src, SPAN_WARNING("Your hand gets pushed away from \the [src]. !"))
+	return
+
+/mob/living/carbon/superior_animal/human/solelite/melee
```
cdb-is-not-good, line 64: You made this a child of the ranged one, i strongly suspect it's going to have a ranged attack if you test it.

code/modules/mob/living/carbon/superior_animal/human/subtype/solelite/solelite
```diff
@@ -0,0 +1,100 @@
...
+	min_air_pressure = 0
+	min_bodytemperature = 0
+
+//Drops
+	meat_amount = 1
+	meat_type = /obj/item/reagent_containers/food/snacks/meat/human
+
+	inherent_mutations = list(MUTATION_HEART, MUTATION_LUNG, MUTATION_LIVER, MUTATION_BLOOD_VESSEL, MUTATION_MUSCLES, MUTATION_NERVES)
+
+	faction = "elitemerc"
```
cdb-is-not-good, line 52: Out of curiosity, why that faction? To piggyback off other functionality?

code/modules/mob/living/carbon/superior_animal/human/subtype/solelite/solelite
```diff
@@ -0,0 +1,100 @@
...
+
+/mob/living/carbon/superior_animal/human/solelite/handle_breath(datum/gas_mixture/breath) //we have are own air supplies
+	return
+
+/mob/living/carbon/superior_animal/human/solelite/handle_environment(var/datum/gas_mixture/environment) //are armor legit is a hardsuit
+	return
+
+/mob/living/carbon/superior_animal/human/solelite/start_pulling(var/atom/movable/AM)
```
cdb-is-not-good, line 60: This is good, I'd also suggest adding a grab sanity check to prevent players from trying to CQC the dude in powered armor. If you need an example of that, look at the render and seekers.

code/modules/mob/living/carbon/superior_animal/human/subtype/solelite/solelite
```diff
@@ -0,0 +1,100 @@
...
+	projectiletype = /obj/item/projectile/bullet/rifle_75/hv
+	drop_items = list(/obj/item/gun/projectile/revolver/hornet)
+	projectilesound = 'sound/weapons/guns/fire/12mm_revolver.ogg'
+	rounds_left = 6
+	mag_type = /obj/item/ammo_magazine/speed_loader_kurtz_50/empty
+	mags_left = 5 //5+1
+
+/mob/living/carbon/superior_animal/human/solelite/shotgunner
+	desc = "Black and charred hardsuit, omnirifle shotgun in-hand, a being of absolute terror to those who dare approach."
+	icon_state = "syndicate_stormtrooper_shotgun"
```
cdb-is-not-good, line 85: This comment applies to all the sprites, honestly - but it might be good to kitbash the wielded sprite of the weapons we want them to have with the sprite they use - these sprites predate Eris iirc and are the old ss13 style of 'guns are only held in one hand always.'

code/modules/mob/living/carbon/superior_animal/human/subtype/solelite/solelite
CactusMouth: No reason to change it, unless we'd want to use both the normal solfed and the elite solfed troops together, though I know you're doing something with them already.

cdb-is-not-good:
> > Are these vatgrowns or meant to be standins for actual humans, because if the latter I'd strongly suggest against using them in the future. If the former, I'm fairly certain the Federation didn't ...use vatgrowns in this way, not that i blame you for making use of what you had available.
> 
> That's the main problem I'm facing. Most of the time you can get NPCs as "Oh they're Vatgrown so they're dumb as rocks." or etc. But this is _Solfed_, and can't do shit like that. Plan is to use them less as NPC mobs and more as easy to spawn puppets during events which ghosts can use, as spawning an entire solfed carbon even if made easier now needs prep.

tbh keep the number of solfed people to a minimum then and just use the suits/spawners to curate them. if each and every solfed agent feels like a character instead of just a throwaway, that's a positive.

cdb-is-not-good:
Closing for now as this has been left idle.

PR #4923, opened 2024-01-11
Taser changes.

cdb-is-not-good:
Taser beam has lost its AP, the combination of hitscan, commonality and pain damage makes me think it really...doesnt deserve AP.

Electrode bolt(such as those used by the counselor and zeus.) which DIDN't have AP have been given the AP that taser beams had, they are slower and a bit harder to use - and giiven that they fit the 'heavy big boy stuns, this seems fiine'.

Stunbolts(as used by the counselors big boy fire mode and shotgun stunslugs) have the above +20 ap, though I mused on giving them extra in light of how limited these shots tend to be, but for now just 20AP.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/energy.dm
+++ b/code/modules/projectiles/projectile/energy.dm
@@ -50,6 +50,7 @@
 	name = "electrode"
 	icon_state = "spark"
 	mob_hit_sound = list('sound/weapons/tase.ogg')
+	armor_penetration = 20
 	nodamage = 1
 	taser_effect = 1
 	agony = 40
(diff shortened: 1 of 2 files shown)


SingingSpock:
Should document that you *did* actually give the Stunshot 40 AP

code/modules/projectiles/projectile/energy.dm
```diff
@@ -60,6 +61,7 @@
 /obj/item/projectile/energy/electrode/stunshot
 	name = "stunshot"
 	damage_types = list(BURN = 5)
+	armor_penetration = 40
```
cdb-is-not-good, line 64:
```suggestion
```

cdb-is-not-good:
Oops, that was not intentional, I thought I undid that but apparently didn't push the change. Fix'd

PR #4925, opened 2024-01-12
Adds the remnant of the ex-solfed base to the deep forest (almost no loot)

Chevalt:
This is mainly for fluff, there was an event including a few ex-solfed soldiers having established a small FOB in the forest, and the colony fought against it. For the sake of persistence, I added the base to the map in its fought-over, sabotaged state. I added spent casings of the types of ammo used in the fight, and one empty magazine, and kept the damage inflicted to it by colonists. There are two very low tier spawners for a gun and ammo, however these will only spawn very cheap and again, low tier weapons, like the makeshift Luty SMG or such just to make it somewhat relevant for the occasional outsider that will wander in. 
The solfed soldiers fled, so the base is in disarray and mostly unuseable. The salvage isn't that good too. 
A few jungle birds have been added to the base, aswell as a mine or two.
solfed base


cdb-is-not-good:
No pictures and I'm strongly tempted to veto this solely on the grounds that it goes against my continued insistence that solfed related content not be used in a casual fashion or with a laissez faire approach.

cdb-is-not-good:
Post some pictures of the thing itself and I'll peek it when I get a chance.

cdb-is-not-good:
personally reviewed, seems fine.

cdb-is-not-good:
seems fine

PR #4938, opened 2024-01-15
Las mg fix

Rat-Time:
Updated the concilium lasMG disk to have large cells instead of medium cells since they were reworked to use large.
Removed a warhammer 40k reference on the concilium lasMG fire modes.


cdb-is-not-good:
Ah nice, good PR

PR #4940, opened 2024-01-15
telesci fixes/buffs

benj8560:
I always felt telesci was way to much of a pain. After some digging I noticed it's not actually been coded to account for are map. Most of this is I would argue a bug fix but the balance implications are still pretty major regardless so test merge!

Changes are as follows

Comments out telescis proc to check for SealedLevel. Since all of are z-levels are sealed this was maxing out the difficulty of telesci regardless of what level/area you target.

What this means in practice: The local colony z-levels should range from a difficulty of 1-4 depending on range, off-colony locations will almost always be 4 or  5, any area with a inhibitor or equivalent will be regardless of range 5. Also telesci I _believe_ can be setup with less relays. You need 1 relay per difficulty level from what I can tell. Meaning SI could do shorter range teleports around the colony with much less mat investment.

Originally regardless of location or map it would max out the difficulty to 5 due to the sealed proc. 

~~I've also mildly bumped the pathfinding speed buff from cognition. a 4 to a 6. Might need to be even higher or not touched at all but I'd like to see this in practice. Tickrate differing so much on my test server makes it hard.~~ Was fun to try out but likely way to stronk in light of how much faster low difficulty teleports are.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


code/modules/telesci/telesci_computer.dm
```diff
@@ -165,8 +165,8 @@
 	var/turf/origin = get_turf(telepad)
 	if(!isPlayerLevel(targetZ))
 		return BS_DISTANCE_INVALID //Invalid area.
-	if(isSealedLevel(targetZ))
-		return BS_DISTANCE_STRESSFUL //Sealed Z levels like deep maint require stressing the telepad.
+//	if(isSealedLevel(targetZ)) //Commented out. Nearly all of are z-levels are sealed  we don't need to check for this. -Benl8561
```
cdb-is-not-good, line 168:
```suggestion
//	if(isSealedLevel(targetZ)) //Commented out. Nearly all of our z-levels are sealed  we don't need to check for this. -Benl8561
```

cdb-is-not-good:
fine by me, we'll testmerge just to amke sure it has no major issues on testing.

SingingSpock:
I heard from others that nullifiers/the SI one didn't properly block inbound teleportation like they're supposed to, could you check if that's the case and maybe see about fixing it if so?

benj8560:
> I heard from others that nullifiers/the SI one didn't properly block inbound teleportation like they're supposed to, could you check if that's the case and maybe see about fixing it if so?

looking at the code they are not suppose to block it its odd. Apparently it just makes it annoying as fuck

benj8560:
more or less like explained vaguely in the PR. It checks if the room has a inhibitor of some type. If it does it maxes out the difficulty. Just due to telesci being ported from a server with a ship map it did not account for the way we tagged are z-levels it was ALWAYS max difficulty so I never noticed this behavior.  I suppose it does behave as advertised. It inhibits teleportation it does not prevent it.

benj8560:
but now that i think on it I might ramp up the amount of damage it does to the relays when you try and do difficulty 5 teleports. Maybe even to the point the teleporter is likely to fail before it runs out of charge to give them more impact in countering it beyond just making it a pain in the ass.

benj8560:
WEll after some more testing its not exactly behaving how I expected it to. It is mostly working. Local colony stuff is ez. But places I expected to be 4 or so outside the colony are getting 5s. Not a super big deal its proly better this way for balance

PR #4948, opened 2024-01-16
Psionic food power now burns rather than takes blood, also reflavored

SingingSpock:
## About The Pull Request
<details>
<summary>
	Makes the psionic food power now give some burns rather than take your blood, changes the amount, and reflavors it
</summary>
<hr>

Psychosomatic Transference is renamed Psychosomatic Fullness, and is flavored as convincing your stomach it's not actually that hungry (which fits the actual meaning of the word psychosomatic). Changed it to give +100 nutrition rather than just setting your nutrition to 400. This is twice as strong as Soul Hunger, but given that 1 psi point is a bigger expenditure than 20 faith and this has a stronger downside I think it's fine. Gave it 10 burn damage (5 with Psi Attunement) rather than costing blood, because the blood cost would almost instantly negate the nutrition gain if you didn't have Attunement. Shouldn't run into the problem Church did with burn damage before, as you shouldn't need to spam this power to do other things with the nutrition.
	
<hr>
</details>

## Changelog
:cl:
tweak: Psionic food power is now called Psychosomatic Fullness and burns you rather than costing so much blood it's useless.
/:cl:

### DIFF
--- a/code/modules/psionics/psion_powers/healing_powers.dm
+++ b/code/modules/psionics/psion_powers/healing_powers.dm
@@ -23,18 +23,18 @@
 //Heals hunger
 /obj/item/organ/internal/psionic_tumor/proc/psychosomatictransfer()
 	set category = "Psionic powers"
-	set name = "Psychosomatic Transference (1)"
-	set desc = "Expend a single point of your psi essence to fill your stomach with cannibalized proteins from your own body. Beware, this will generate toxins and expend some of your blood."
+	set name = "Psychosomatic Fullness (1)"
+	set desc = "Expend a single point of your psi essence to convince your stomach it's not actually that hungry, burning fat reserves to keep going strong. Taxing on the mind and causes minor burns."
 	psi_point_cost = 1
 
 	if(pay_power_cost(psi_point_cost))
 		if(!owner.stats.getPerk(PERK_PSI_ATTUNEMENT))
-			owner.nutrition = 400
-			owner.drip_blood(54)
+			owner.nutrition += 100 //Twice as strong as Soul Hunger
+			owner.adjustFireLoss(10) //You're not using your nutrition to fuel things like Church used to, this should be fine
 		else
-			owner.nutrition = 400
-			owner.drip_blood(26)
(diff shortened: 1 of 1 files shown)


Trilbyspaceclone:
This makes it basically free as you naturally will heal back those damage values in like 2 mins

benj8560:
> This makes it basically free as you naturally will heal back those damage values in like 2 mins

As it is now it uses up so much blood you end up starving. There is some way to work around it like liquid rations with iron but then you might as well just eat the food. :P

SingingSpock:
> This makes it basically free as you naturally will heal back those damage values in like 2 mins

Takes about 20 minutes, and that's assuming you don't get hit at all in the meantime. And as Ben says, the current version is quite literally useless unless you have the perk, as the bloodloss will drain you back down to starving.

SingingSpock:
> burns heal _about_ 0.05 every natural healing it should not take 20 mins it should like normally tool mess ups take 2 mins

That's 200 natural healing cycles to heal a 10 damage burn. This is also just... doubling what Soul Hunger used to be, both in effect size and damage.

Trilbyspaceclone:
Im just stating that it seems it dosnt have a real drawback now

cdb-is-not-good:
5 fireloss is *REALLY* negligible, i agree. I suggest picking something that isn't quite....so nothing.

AysheDaArtist:
People drink sugar, people spit out ERP meals from their borgs, it's a small thing if Psionics eat a little psi-points and burns, otherwise pretty shit power ngl without these changes

SingingSpock:
5 fireloss is only if you have the specific perk that's supposed to make your psionics have a lower cost. That requires a specific career that doesn't give any stats, a fair trade imo. If you don't have that, it's 10 fireloss.On Jan 16, 2024, at 20:35, CDB ***@***.***> wrote:﻿
5 fireloss is REALLY negligible, i agree. I suggest picking something that isn't quite....so nothing.

cdb-is-not-good:
> People drink sugar, people spit out ERP meals from their borgs, it's a small thing if Psionics eat a little psi-points and burns, otherwise pretty shit power ngl without these changes

I suppose that's a fair enough point.

PR #4951, opened 2024-01-18
RIG weirdness, star trader and objective bug fixs

benj8560:
starting with the small stuff: 
the greysons Star traders vender no longer gives out free **genuine** greyson coats. Now giving the star traders version at cost.

The guild APC disturbance objective now works. Trilby helped me with this one thanks. The one where you need to go power down a APC in a area (mostly). There is still a issue related to this objective selecting areas that don't exist on map. I will try and fix this later but the files I think are the problem are locked up currently. 

RIG stuff:

Goes through I believe every single RIG and standardizes their naming there should be no more weirdness of a ''hussar RIG hussar hardsuit boots'' across all RIGs. I did not test every single suit but it _should_ be fine I checked a fair few. 

The nanite RIG has been giving 100 rad protection this was meant to be done awhile back. I missed it since its hiding in its own file away from other RIGs.

Not changelog related note to coders: PLEASE if you make a RIGsuit. Follow the proper standard for this. Just lookie at the code its very simple! The Proto-Spacer is a good example if you wanna give the helmet/whatever a fancy name. Same standard all the older RIGs use. Dunno when this weird trend started on soj.

### DIFF
--- a/code/game/machinery/vendor/one_star_vending.dm
+++ b/code/game/machinery/vendor/one_star_vending.dm
@@ -89,7 +89,7 @@
 	products = list(
 		/obj/item/clothing/mask/smokable/cigarette/os/nova = 25,
 		/obj/item/clothing/suit/space/void/os/nova = 15,
-		/obj/item/clothing/suit/greatcoat/os = 10,
+		/obj/item/clothing/suit/greatcoat/os/nova = 10,
 		/obj/item/clothing/under/os_jumpsuit/nova = 10,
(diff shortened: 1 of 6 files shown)


cdb-is-not-good:
>Not changelog related note to coders: PLEASE if you make a RIGsuit. Follow the proper standard for this. Just lookie at the code its very simple! The Proto-Spacer is a good example if you wanna give the helmet/whatever a fancy name. Same standard all the older RIGs use. Dunno when this weird trend started on soj.

image

yes maam

PR #4952, opened 2024-01-18
Mapping in what the colonists build (event reward)

Chevalt:
An event has happened in which the colony was tasked with building a few outposts, they only renovated the spider outpost partially, and blew up their own roads and buildings in the process. I have mapped what they built and did, that is all. This contains the "renovated" spider outpost and the craters the colony did by maxcapping the place.
image
image


SingingSpock:
Okay, so the question I have is why there aren't *any* spawns for hostiles inside the outpost. It's not the Spider outpost because the spiders built it, but because the Colony can't consistently project enough power past it's borders (or even within its own walls) to actually keep it clear. Also removing the loot from it removes one of the few non-claimed sources of decent loot on the Colony map.

Chevalt:
> Okay, so the question I have is why there aren't _any_ spawns for hostiles inside the outpost. It's not the Spider outpost because the spiders built it, but because the Colony can't consistently project enough power past its borders (or even within its own walls) to actually keep it clear. Also removing the loot from it removes one of the few non-claimed sources of decent loot on the Colony map.

Because they literally just built it yesterday, so I'm leaving a bit of time before a second event where they'll fix their mistakes. In the meantime, I'll probably map in mobs and degradation in a week or so, to simulate the jungle taking over it. There is also no loot because.. well, that's how it was when they built it, no stockpiles of anything, no ressources. I just delivered on the promise I made when I decided to make the event, I even asked in staff lobby if people were fine with it.

Chevalt:
> > Okay, so the question I have is why there aren't _any_ spawns for hostiles inside the outpost. It's not the Spider outpost because the spiders built it, but because the Colony can't consistently project enough power past its borders (or even within its own walls) to actually keep it clear. Also removing the loot from it removes one of the few non-claimed sources of decent loot on the Colony map.
> 
> Because they literally just built it yesterday, so I'm leaving a bit of time before a second event where they'll fix their mistakes. In the meantime, I'll probably map in mobs and degradation in a week or so, to simulate the jungle taking over it. There is also no loot because.. well, that's how it was when they built it, no stockpiles of anything, no ressources. I just delivered on the promise I made when I decided to make the event, I even asked in staff lobby if people were fine with it.

Oh and they also maxcapped the spider outpost twice.. So.. I mean, it wouldn't also make any sense to leave the old loot in, considering it essentially suffered 5 kilotons of TNT?

benj8560:
wait they leveled the visitors cabin? Make sure you yell at them to fix it when event time comes I liked that thing was neat for hanging out near the colony as outsider

Chevalt:
> wait they leveled the visitors cabin? Make sure you yell at them to fix it when event time comes I liked that thing was neat for hanging out near the colony as outsider

I'll make sure they at least try to fix what they blew up in the next event.

PR #4966, opened 2024-01-23
church melee balance

benj8560:
long needed look at of church melee weapons. I may do more I may call it good here. Input is welcome from anyone in terms of what needs changing.

So far changes are as follows:

Both church belts can now hold the church shortsword. Making it have a place as a effective backup melee.

The warhammer now has bonus structural damage idk why it never had such.

The NT spear has had its god damn 50! AP reduced to 30 making it no longer the best PVE church melee.

The Vexilar force blade has had its active time doubled. The most expensive church melee should likely not just be a gimmick. Around 12 seconds or so but mileage may vary with tickrate as usual.

### DIFF
--- a/code/game/objects/items/weapons/nt_melee.dm
+++ b/code/game/objects/items/weapons/nt_melee.dm
@@ -140,7 +140,7 @@
 	w_class = ITEM_SIZE_HUGE
 	slot_flags = SLOT_BACK | SLOT_BELT
 	throwforce = WEAPON_FORCE_LETHAL * 1.5
-	armor_penetration = ARMOR_PEN_HALF
+	armor_penetration = ARMOR_PEN_MASSIVE
 	throw_speed = 3
 	price_tag = 150
 	matter = list(MATERIAL_BIOMATTER = 20, MATERIAL_PLASTEEL = 10) // More expensive, high-end spear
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/nt_melee.dm
```diff
@@ -266,7 +267,7 @@
 	switched_on_qualities = list(QUALITY_CUTTING = 30, QUALITY_SAWING = 30)
 	switched_off_qualities = list(QUALITY_CUTTING = 10, QUALITY_SAWING = 10)
 	tool_qualities = list(QUALITY_CUTTING = 10, QUALITY_SAWING = 10)
-	active_time = 50
+	active_time = 100
```
SingingSpock, line 270: Time is in deciseconds, which is why leaving it unmodified is bad. 10 SECONDS would be better (the all caps is important)

benj8560, line 270:
```suggestion
	active_time = 10 SECONDS
```
Like so?

SingingSpock: Yes

benj8560:
this...was suppose to be test merged I figured it was obvious, Balance changes!

PR #4975, opened 2024-01-27
More church gun balance and small fixes.

benj8560:
Bug fixes first: Fixes the Bat'ko grabbing its serial number from the lightfall
Fixes many church weapons mislabeling themselves as NT (new testament) and vice versa for OT (old testament). inside autolathes. Removes the lightfall from the OT disk. It was on both idk why. Lives on NT disk only now.

Important balance stuff: Concillium has seen well needed nerfs/fixes from my last time touching it. Fixes it incorrectly having 1 fire delay if you never change the fire mode. 
Pretty substantially reduces it's fire rate. I never wanted this to be a DPS monster and somehow it ended up that way.
It's recoil is now a factor. By no means is it bad but if you want it to be recoilless you'll be modding it or bracing it. Should discourage running and gunning as well
Ups its gold cost even more. Really trying to avoid slapping a uranium cost onto this one.

slightly bumps the cost of the Dominion from 2 to 2.5 (in bioprinters). Apparently? It was just cheap enough if you took a loadout wrist PDA for a extra atom cell you could print one of these nearly round start.

After many requests Lightfall has been buffed a fair amount. Halfing its charge cost. It's now comparable to a better fire rate cog (aside from in mat costs) Silver cost up to 10 as well (5 in bioprinter) and it's size was changed to bulky. No funny pocket rifle!

### DIFF
--- a/code/datums/autolathe/guns.dm
+++ b/code/datums/autolathe/guns.dm
@@ -445,7 +445,7 @@
 	build_path = /obj/item/gun/energy/gun/martin
 
 /datum/design/autolathe/gun/ntpistol
-	name = "OT LP \"Serenity\""
+	name = "NT LP \"Serenity\""
 	build_path = /obj/item/gun/energy/ntpistol
 
 /datum/design/autolathe/gun/lasercore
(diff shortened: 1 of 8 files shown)


benj8560:
also would like to point out the outright silly fact that the hand-made laser carbine is virtually identical to the lightfall aside from a TINY bit higher fire delay prior to this PR. Truly silly.

cdb-is-not-good:
gonn a test merge just to make sure there's no balance issues that result

PR #4980, opened 2024-01-29
Adds Reaper to the spider deep-maints entrance

SingingSpock:
## About The Pull Request
<details>
<summary>
	Adds one reaper spider to the deep-maints entrance guarded by spiders
</summary>
<hr>

image
Removed one emperor trap and one low_chance trap spawner that were rather close to the entrance to make it less ultradeadly. Still dangerous because the Reaper is big, fast, and has sleepy toxin.
<hr>
</details>

## Changelog
:cl:
map: Adds a reaper spider to the deep maints entrance guarded by spiders
/:cl:


ChefDoggo:
reapers are basically meant to be only admin spawn, dunno about why this would get merged unless the reaper actually gets a revamp to be a threat

SingingSpock:
I mean, if it's not a dangerous threat I don't see why adding it will be a problem. It's definitely a greater threat than the single emperor and maybe trapdoor spider that it'll replace.

cdb-is-not-good:
I will probably want to push a PR to make reapers more befitting of their leviathan title if we're going to have one just placed in the world rather than being an admin only mob that gets used for events and usually with tweaking. That or recategorize them out of the leviathan class along with kaisers

SingingSpock:
Okay, so I didn't understand the state of Reapers or where they _should_ be after getting buffed to something reasonable. I'm closing this one, may make one in the future with one in the swamps.

PR #4981, opened 2024-01-29
Stingers and bouncing bullets.

cdb-is-not-good:
Stinger grenades will no longer doo 400+ damage to targets they're directly under/on top of, instead being closer to 100. Stingers also have a higher damage_step comparable to eris stingers- this means the projectiles travel more and will bounce about more. These changes should hopefuly make stingers less of a crapshoot between 'literally a joke' and 'Stronger than a fragmentation grenade'. Stingers also can ricochet now

Additionally, implements a new variable that modifies how likely a bullet is to ricochet. This variable is a 1 by default and is negatively affected by armor pen. Lower cal rounds are more likely to bounce(about 3x as likely for 9mm vs .408 which has a net bonus of 1.15x). This also affects stingers, which have a comparable ricochet chance to 9mm, this is intended to make them more useful outside of point blank range.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


cdb-is-not-good:
Tested and seems to work just fiine, it's not AS dramatic as I'd like but that's fine for now. Can test it and see if it does what we want.

PR #4985, opened 2024-01-30
Bullits 'n woundiing mods.

cdb-is-not-good:
## About The Pull Request
<details>
<summary>
Actually adds wounding mult from Eris and makes it a bity more granular(i felt this necessary due to our uh...frankly insane amount of weapons/guns.) - For the record, I only slightly understand how this actually ineracts with the code that Rebel decided to port that referenced it but I THINK it affects how likely (specifically bullets) are to do organ damage in addition to actual regular damage. I spread the mults across our ammo in a roughly homogenous way but we'll need to test to see if it like....matters. If i understand right, this system will mean 9mm HV will tend to wound less than 9mm hp, which will do worse than 10mm hp and so on.

Also adds a tweak to the damage proc to *maybe* fix weirdness owing to us not using armor divisor, instead generating armor divisor by dividng AP by 10(so you still end up roughly....the same scale of 1-10ish).

Also nukes PVP/PVE damage listing, damage is now given only as 1 value that incidentally also includes modifiers for post_pen_damagemult(HP damage bonus).  PLEASE note that the damage listed for HALLOWPOINT bullets and other TRAUMATIC ROUNDS(Shrapnel bolts, red-nose AMR, buckshot) is HIGHLY SUBJECT to ARMOR VALUES and if you use it against armored targets it will NOT WORK GREAT. this has not changed, but I want to make it double clear before this goes live and people scream 'cdb you broke hp awaawawawawa'
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


cdb-is-not-good:
whew boy that's a lotta commit junk. good thing I don't care and we squash these.

PR #4987, opened 2024-01-31
Sensible Benzin roaches

benj8560:
A near direct port of benzins from https://github.com/Eclipse-Station/NEV-Northern-Light. Most of the changes are under the hood but some big mechanical things to note:

Most importantly very likely fixes a few major runtimes related to these annoying fellows tho time will tell.

Benzins will now explode based on projectile. Rubbers, arrows, etc _should_ not blow them up. Cant say I got every example for sure but I SHOULD have. Common sense should apply here for the most part. But most projectiles will explody.
Doing ANY form of burn damage to a benzin will make it explode. Laser scalpels, deep maint weapons, etc, If it burns it blows.
Benzins now only explode in a small 3x3 area. Smol. Less pain.
Likely other things I'm forgetting. Pretty much if something seems like it will blow one up it will! (including after death)

Using a sharp weapon of any kind to attack a benzin will trigger a leak. It will leak a large amount of welding fuel when dragged making a trail.

### DIFF
--- a/code/game/objects/items/weapons/tools/_tools.dm
+++ b/code/game/objects/items/weapons/tools/_tools.dm
@@ -1010,8 +1010,8 @@
 				var/obj/item/weldpack/P = O
 				P.explode()
 			return
-		else if(istype(O, /mob/living/carbon/superior_animal/roach/benzin))
-			var/mob/living/carbon/superior_animal/roach/benzin/B = O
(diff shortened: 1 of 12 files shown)


cdb-is-not-good:
Be sure to credit the actual PR or coder in the future.

PR #4988, opened 2024-01-31
Tisane description fix

WilsonWeave:
Fixes the description of mercys hand tisane to accurately reference itself instead of two separate wrong fruit.

Image to show the before and after results, and that it compiles/launches.

image


cdb-is-not-good:
wow damn, how'd i fuck that up so b ad. Thanks Wilsong.

PR #5000, opened 2024-02-01
Replaces the Spider Outpost with an underground Spider Cave

SingingSpock:
## About The Pull Request
<details>
<summary>
	Spider Outpost collapses into the ground, exposing the spider cave that was beneath it!
</summary>
<hr>

Screenshot 2024-02-01 152941

Screenshot 2024-02-01 152956

Replaces the spider outpost with a spider cave, which had sometimes been said to be beneath it in the past. This cave is on the Underground 1 level, and has some mineable rock that leads to the colony's maintenances. This gives the colony a second entry point via mining into maints, which is good for allowing hostile outsiders to actually do something other than die in the meadow as they fail to get into the colony proper.

I tried to balance the loot and loot types against the amount of mobs, but it might be a bit off. I tried to err on the side of less rather than the side of more.
<hr>
</details>

## Changelog
:cl:
map: Replaces the Spider Outpost with an underground Spider Cave
/:cl:

### DIFF
--- a/maps/__Nadezhda/area/_Nadezhda_areas.dm
+++ b/maps/__Nadezhda/area/_Nadezhda_areas.dm
@@ -311,6 +311,7 @@
 /area/nadezhda/outside/scave
 	name = "Spider Cave"
 	icon_state = "erisgreen"
+	dynamic_lighting = TRUE
 
 /area/nadezhda/outside/forest
 	name = "Forest"
--- a/maps/__Nadezhda/map/_Nadezhda_Colony_New.dmm
+++ b/maps/__Nadezhda/map/_Nadezhda_Colony_New.dmm
@@ -17635,9 +17635,9 @@
 /turf/simulated/floor/tiled/dark/brown_perforated,
 /area/nadezhda/storage/primary)
 "dOt" = (
-/obj/random/gun_shotgun,
-/turf/simulated/floor/asteroid/dirt/dust,
-/area/nadezhda/outside/scave)
+/obj/item/spider_shadow_trap,
+/turf/simulated/floor/plating,
+/area/nadezhda/dungeon/outside/abandoned_outpost)
 "dOv" = (
 /obj/structure/cable/green{
 	d1 = 4;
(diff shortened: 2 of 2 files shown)


SingingSpock:
Accidentally undid someone else's changes, let me see if I can fix that. Don't merge yet.

SingingSpock:
There we go

cdb-is-not-good:
Giving until Monday for Chev and co to properly resolve the event they did halfway, if they dont by then i'll merge this.

Support-hells-angels-bitch (changes requested):
Listen linda

Support-hells-angels-bitch:
Listen Linda Reddog wants to talk with Tracy, checkmate!!

cdb-is-not-good:
I may ask you to make some tweaks to this, but i wanna see it in play first. With this new, fairly larger breach now present I might suggest bolting the doors and stapling a 'DO NOT TRESSPASS' sign on the outside as well as  either offering better camera coverage of the east/west mountain soft spots or simply removing those entryways since there's now a far better way in/out(illegal style). Mostly just so the colony isn't a hilarious sieve in terms of security holes

cdb-is-not-good:
Right off the bat, those rare gun spawns are a BIT much for an area that's just got a few spiders in :p

SingingSpock:
> Right off the bat, those rare gun spawns are a BIT much for an area that's just got a few spiders in :p Specifically, the energy_weapon spawn at the south-east corner, the dungeon_energy_gun spawn at the north east. AFAIK the spider outpost didn't have any spawns that weren't cheap spawners.

I'll get on that in the morning. I don't actually know what guns are in each spawn, it isn't really well labeled how good each one is.

cdb-is-not-good:
Think this is good for now, I've a few tweaks I wanna make but that's probably better suited to its own PR due to *what* i wanna do. In the future for these kinds of over-hanging thiings I'd probably suggest using latice+catwalk(after all, it's cheaper to produce :p)

PR #5019, opened 2024-02-07
buffs Entreaty

benj8560:
Makes the common cruciform litany Entreaty give exact coordinates as well as area.  I simply believe the most common places this is useful have far to big a area to not provide coordinates. Junk fields, maint, dungeons etc.  ~~This is pretty difficult to test on a private server but it mimics the way death alarms handle it so SHOULD work.~~ Tested. Works.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/rituals/base.dm
+++ b/code/modules/core_implant/cruciform/rituals/base.dm
@@ -92,9 +92,10 @@
 
 		var/obj/item/implant/core_implant/cruciform/CI = target.get_core_implant()
 		var/area/t = get_area(H)
+		var/turf/T = get_turf(H)
 
 		if((istype(CI)))
-			to_chat(target, SPAN_DANGER("[H], faithful cruciform follower, cries for salvation at [t.name]!"))
+			to_chat(target, SPAN_DANGER("[H], faithful cruciform follower, cries for salvation at [t.name] [T.x], [T.y], [T.z]!"))
 	return TRUE
 
 /datum/ritual/cruciform/base/reveal


benj8560:
test merge suggested due to the untested code thing.

SingingSpock:
I think this is an unnecessary buff to a litany that's already powerful if used well. Churchies are able to get death implants for precise death locations. Entreaty is useable while alive, does not depend at all on radio, and goes only to other Absolutists. Adding precise location information makes it even harder to do various crimes like hostage taking, and would also allow it to be used for long-distance communication of precise locations for other purposes outside its design intent.

benj8560:
tested on live and seems to work as intended.

cdb-is-not-good:
I kiiiinda agree with spock on the grounds of balance, like. does it *need* a buff?

benj8560:
> I kiiiinda agree with spock on the grounds of balance, like. does it _need_ a buff?

Its kinda good. In very  specific circumstances. But....I still don't believe it really changes much. Like I said it's genuinely a downgrade to all other choices in most of the circumstances spock stated. The most practical place for this to be used is in a long list of z-levels that are MASSIVE rendering it more annoying to the players in game trying to help then it's worth. I'd also like to note how underwhelming the small list of litany's non-vector church have. This would at least give some use to those people. just LOOK at the other implants with major mechanical gain over this.

 I could compromise by raising the charge cost? It is sending a potentially long list of people a message. It seems reasonable it might use a lot of charge.

SingingSpock:
> > I kiiiinda agree with spock on the grounds of balance, like. does it _need_ a buff?
> 
> Its kinda good. In very specific circumstances. But....I still don't believe it really changes much. Like I said it's genuinely a downgrade to all other choices in most of the circumstances spock stated. The most practical place for this to be used is in a long list of areas that are MASSIVE rendering it more annoying to the players in game trying to help then it's worth. I'd also like to note how underwhelming the small list of litany's non-vector church have. This would at least give some use to those people. just LOOK at the other implants with major mechanical gain over this.
> 
> I could compromise by raising the charge cost? It is sending a potentially long list of people a message. It seems reasonable it might use a lot of charge.

The cost needs to stay low so it can perform its actual purpose of being a panic button that lets everyone know you're in danger, with a general location. Yes, it isn't all that useful outside the colony in the big open areas, and quite honestly _that's fine_. If you're out there, you're either an outsider or going adventuring, you get less protections. We shouldn't balance this around that, especially since, as I've pointed out, _death alarms are cheap, available to Churchies, and give the exact coordinates_. And yes, the base-level cruciform isn't all that powerful, that's just how the balance goes... part of it is that you augment the basic litanies with your Path litanies to make a package that's better. Not as good as base psionics, but it also doesn't make you take +20% damage or lock you out from implants.

cdb-is-not-good:
> > I kiiiinda agree with spock on the grounds of balance, like. does it _need_ a buff?
> 
> Its kinda good. In very specific circumstances. But....I still don't believe it really changes much. Like I said it's genuinely a downgrade to all other choices in most of the circumstances spock stated. The most practical place for this to be used is in a long list of areas that are MASSIVE rendering it more annoying to the players in game trying to help then it's worth. I'd also like to note how underwhelming the small list of litany's non-vector church have. This would at least give some use to those people. just LOOK at the other implants with major mechanical gain over this.
> 
> I could compromise by raising the charge cost? It is sending a potentially long list of people a message. It seems reasonable it might use a lot of charge.

Jack of all trades, master of none. That is and has always been how we want the church balancing to look - so saying 'its just not quite as good as other options' ...makes me question if this PR is really necessary.

benj8560:
Fair arguments before we decided to add a mechanical downside (a pretty annoying one at that) to cruciform. Its far more annoying then its worth to take one outside of vector. I'd personally like to see more church just around the colony being part of the gang. Like 70% of players have a gate or psi organ where almost none outside of vectors have a cruciform. I feel like this should say enough? Your lucky to see even 1 per round.

benj8560:
this was the case before we even added a mechanical downside

cdb-is-not-good:
> Fair arguments before we decided to add a mechanical downside (a pretty annoying one at that) to cruciform. Its far more annoying then its worth to take one outside of vector. I'd personally like to see more church just around the colony being part of the gang. Like 70% of players have a gate or psi organ where almost none outside of vectors have a cruciform. I feel like this should say enough? Your lucky to see even 1 per round.

what mechanical downside would that be? Even regardless of that, the cruciform remains the most powerful and versatile of all the core implants and the major "downside" is that it's tied to a faction/religion, I'm not really hearing an argument against "why should this thing function as well as other options".

PR #5022, opened 2024-02-07
Some tweaks to the spider cave.

cdb-is-not-good:
The Blackshield and Guild have finally gotten around to fencing off the terrible pit that is the spider-cave in the ruins of the old spider pit - unfortunately, it has sunk even further down and seems to have finally become stable. New safety measures have been implemented in line with OSHA compliance.

Spider pit goes all the way down instead of just one layer down and instead of opening up to a random maints tunnel either can be tunneled through to the render cave or the maints medbay. Generally attempts to add some polish to the area without making it too impossible to use it to sneak in.
image
image
image

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


cdb-is-not-good:
i hate code

Issue #4866, opened 2023-11-25
rad resistance gene not always working


benj8560:
found so far two cases of the rad resist xenogenetics gene not working. Currently Kaisers being the primary issue their bile and possibly the other way they make rads. Not tested super throughly but SOMETHING they do is doing it.

Little more niche of a issue rads from Xenoarch anomalies seem to cause issue as well.

Andron1219:
This also extends to Cindarites not being radiation immune anymore (which they should be immune), probably still using the old radiation damage system.

benj8560:
fixed

PR #5044, opened 2024-02-14
radiation mod fix

Dimasw99:
## About The Pull Request
Radiation mod now applies to the recieved value as well as the damage recieved.

Both the gene and cindarites were no longer immune to radiation because the majority of bad effects and damage procs were moved before the actual damage reduction, likewise not counting in the other nasty effects.

Fixes issue: #4866


benj8560:
So you know should be typed out like '' fixes #4866 '' and it will automatically link the issue. As well as automatically close said issue once merged.

PR #5054, opened 2024-02-18
revert auretian nerf

Dimasw99:
## About The Pull Request
The auretian was nerfed in PR #4552 by a factor of 2.5 which put it down from 20 shots down to 8.
I believe that this was an accidential nerf because the auretian is using a small cell and not a medium cell which most other guns in that PR were balanced around.

The auretian is a Soteria exclusive, high cost and research heavy weapon that is ment to act as a side arm. At the current state the sidearm is just more than useless now, it was fine before. It wasn't the greatest side arm but still a very good one to have and very much worth its investment. Not it is sadly just pure trash and a classica is more usuable than it

Edit: Fixed recoil: Auretian no longer has rifle recoil and instead handgun while the centurio recieved a recoil nerf.
The blacklist restriction that was inherented from the plasma rifle was lifted since the auretian is not full automatic, unlike the church plasma guns

### DIFF
--- a/code/modules/projectiles/guns/energy/plasma/centurio_auretian.dm
+++ b/code/modules/projectiles/guns/energy/plasma/centurio_auretian.dm
@@ -12,6 +12,7 @@
 	matter = list(MATERIAL_STEEL = 20, MATERIAL_PLASMA = 5)
 	price_tag = 1250
 	damage_multiplier = 1
+	init_recoil = HANDGUN_RECOIL(0.2)
 	init_firemodes = list(
 		list(mode_name="melt", mode_desc="Hard hitting plasma bolt that melts flesh and armor alike", projectile_type=/obj/item/projectile/plasma/heavy, fire_sound='sound/weapons/energy/incinerate.ogg', fire_delay = 9, icon="destroy", projectile_color = "#006633"),
 		list(mode_name="ion shot", mode_desc="An iodizing shot to disable cells, electronics and cybernetics ", projectile_type=/obj/item/projectile/ion, fire_sound='sound/effects/supermatter.ogg', fire_delay = 25, icon="stun", projectile_color = "#ff7f24"),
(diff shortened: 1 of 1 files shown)


Dimasw99:
Turned it into a draft because I realized that the recoil balance is fucked

benj8560:
did infact not intend to pass down a unmaker block when I touched plasma weapons. This pleases me 1+

cdb-is-not-good:
Unmaker SHOULD be blocked from most weapons for balance reasons. This is a few too many buffs all at once for me to be confident in merging it.

Dimasw99:
There are much better options to add an unmaker onto a gun. Weapons that benefit the most from an unmaker are those with full auto and a very high fire rate. The auretian suffers from click delays on the other hand and has thus a hard limit on its fire rate and the amount of damage it can do. It is by all means exactly the same as the cog in that regard then.

benj8560:
> Unmaker SHOULD be blocked from most weapons for balance reasons. This is a few too many buffs all at once for me to be confident in merging it.

it only has a unmaker block due to grabbing its unmaker block from the Dominion a block I added and did not intend to pass down as I said. Yes there is definitely several other better choices for slapping as unmaker on no biggy

cdb-is-not-good:
other things being relatively unbalanced isn't in my opinion a good defense to 'this change would be a bad idea for balance'. Wasn't it decided more or less that the unmaker should really only be going on greyson guns or something to that effect

Dimasw99:
> other things being relatively unbalanced isn't in my opinion a good defense to 'this change would be a bad idea for balance'. Wasn't it decided more or less that the unmaker should really only be going on greyson guns or something to that effect

You never really specified the exact balance reason you have in mind because an auretian is the most basic gun you can imagine. It is a cog that you can holster that costs a few extra exotic resources. With its main advantage being able to be holstered instead. Matter of fact even is that the pinion, a greyson weapon, is unable to even hold an unmaker. The only gun that makes the greyson rule fit in RnD is just the cog. As a reminder, the unmaker requires both power and the GP allowed condition so even all the stupid unique and hard to get nail guns and so on from greyson are just out of the option.

In the end it is just another stupid and rather depressing removal from science, similar with the nanoAI being nerfed to shit.
I will undo the unmaker change but I still think that there is a better option to adjusting its balance for PVP (PVE is fine).

benj8560:
> other things being relatively unbalanced isn't in my opinion a good defense to 'this change would be a bad idea for balance'. Wasn't it decided more or less that the unmaker should really only be going on greyson guns or something to that effect


Personally if this gun having a unmaker is a problem? Someones gonna need to go and hit 70% of the e-guns in the game at *least*. It's not all that good. (personally I'd still rather we leave unmaker juiced as fuck and just make it go along side explosive tier contraband if anything)

 It also very expensive (relative to all the other choices R&D has). All around even with this PR its a more expensive downgrade to the Centurio not even a side grade.  Centurio can manage double the capacity and has hands down the best projectile energy ion in the game. It was never intended to have a unmaker block. This is a bugfix. This guns only real advantage is using small cells. But its NEXT most comparable energy the  Centurio has 2x the capacity (round about) medium cells are 2x the cost. 

TLDR theres no point in a unmaker block on this when this argument applies to 70% of the e-guns in the game

benj8560:
> other things being relatively unbalanced isn't in my opinion a good defense to 'this change would be a bad idea for balance'. Wasn't it decided more or less that the unmaker should really only be going on greyson guns or something to that effect

and also no. An unmaker can go on any gun that gets a fancy unmaker allowed var or something. You first nanoAI the gun. THEN the nanoAI can allow for a unmaker (meaning you need energy firing). Greyson guns just have this by default and some guns specifically block it.

PR #5055, opened 2024-02-18
Adds Journalist alt title

Bamhalazam:
## About The Pull Request
Exactly what it says on the tin. A person suggested it to me and seeing as how I am planning to re-do the lower council halls, I thought it would not be a bad idea.

This is an alt title for colonist.


cdb-is-not-good:
sure, why not.

PR #5057, opened 2024-02-18
Update eyes.dm

benj8560:
Burns on the eyes will now progress to their own unique wound types that act  similar to infection slowly killing the organ. It will not however spread to others. You can and still will go blind after welding without goggles but it should be awhile before its a problem.


Dimasw99:
Ah yes, I have welded a spider web. Now I will die from total organ failure as my organs rot from an infection.

benj8560:
Working now ready to test merge. I *desire* a test merge since I've one other idea for how to handle this I might go for depending if this is still whined about a lot

benj8560:
honestly I'd just full merge this. I don't see me collecting the info I need on it in any timely manner. I'll come back to it down the road if needed but currently everyone has PTSD from infection welders and is pretty thoroughly avoiding it at all costs.

PR #5059, opened 2024-02-20
MAYBE fixes issues with pain being insane

cdb-is-not-good:
My testing seems to indicate that the maths that are meant to return only the organ efficiency for 1 limb are actually returning a far larger value resulting in the crazy ass numbers that we see when say....Aulvae/Exalts take pain damage. The maths now should account for the fact that it's TOTAL nerve efficiency, not just one limb. Obviously? Adding even one more nerve will cause significant extra pain.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/shock.dm
+++ b/code/modules/mob/living/carbon/shock.dm
@@ -51,7 +51,7 @@
 		. += organ.internal_wound_hal_dam
 		if(organ && (organ.is_broken() || (!BP_IS_ROBOTIC(organ) && organ.open)))
 			. += 25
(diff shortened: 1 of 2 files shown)


Rat-Time:
Slimes need this very badly.  As it stands they can't take a hit, even one from a slipped tool check, without falling unconscious for a few minutes.  Space shark bite and crovel to the chest both resulted in a KO, wearing a plate carrier and with BS trooper starting stats.

cdb-is-not-good:
better implimentation by dimas, closin.

PR #5067, opened 2024-02-22
Fixes AK frame grips

Rat-Time:
AK frame changed to no longer accept the serb (bakelite) grip as the recipe for /obj/item/gun/projectile/automatic/ak47/sa has changed to wood grips.


benj8560:
Need to fix the other end of it. Allowing it to take wooden grips. Changing it to wooden grips was a intended change just something musta went wrong. The material costs is wood in a lathe, the icon is _visually_ made of wood etc etc

Rat-Time:
Unless I'm just tired and missing something then that part was already done in one of the previous gun changes by Trilby, the serb grip wasn't changed over to wood however which is what this fix does

PR #5068, opened 2024-02-23
Hair Addition: Tri Bun + Jill Bugfix

OliOliOnsiPree:
## About The Pull Request
What this PR does is adds a new hair in from TG that is called Tri Bun and adds it onto here as I plan on using it as my one character, so thats pretty selfish of me in my own right but it is what it is. Thank you Raptor for giving the hair to blackscale Also fixes the Jill hair to actually appear instead of ERROR. Yes no one reported it as I just noticed it now. 
image

## Changelog
:cl:
add: Tri Bun
Fix: Jill
/:cl:


cdb-is-not-good:
👍

PR #5075, opened 2024-02-26
Map Addition: Hunt Master has Gate access

OliOliOnsiPree:
## About The Pull Request
This PR gives the hunt master gate access to freely come into the colony as they are allied with the colony only the hunt master has access.
image
Only these two doors in the gate. Rest is all a IC issue. 

	
<hr>
</details>

## Changelog
:cl:
tweak: gate access with hunt master.
/:cl:


OliOliOnsiPree:
I just realized I did a bad on this PR did the wrong one in here thinking it was right. Whoops.

PR #5076, opened 2024-02-26
Map Addition: Hunt Master has Gate access

OliOliOnsiPree:
## About The Pull Request
This PR gives the hunt master gate access to freely come into the colony as they are allied with the colony only the hunt master has access.
image
Only these two doors in the gate. Rest is all a IC issue. 

Remade of this pr https://github.com/sojourn-13/sojourn-station/pull/5075 but it actually has the right things in it.

	
<hr>
</details>

## Changelog
:cl:
tweak: gate access with hunt master.
/:cl:


cdb-is-not-good:
👍

PR #5091, opened 2024-03-02
Maxhp via leveling cap

LongSleeves:
Adds in a cap for gaining max health thru leveling. Cap is currently set to 300 and can be easily adjusted if needed in the future.


AysheDaArtist:
Absolute KING! ❤

PR #5095, opened 2024-03-04
Language Adjustment: Opifexee

OliOliOnsiPree:
## About The Pull Request
What this PR does is:
Brings the Opifex in line with old "Opifex" by bringing back there old syllables, which is vox, as the language states before it is basically vox, but changes the color to soft red. That way it isnt too similar to Cultist and people dont get confused between the two.;

Color has changed to red on suggested by Trilby
image
image

Rcolor has been changed from Marqua
image
image

Color: 
https://www.colorhexa.com/e96443
https://www.colorhexa.com/5ec2e7
## Changelog
:cl:
add: Red Opifex, Blue Marqua and "new" syllables for opifex
/:cl:

### DIFF
--- a/code/modules/goonchat/browserassets/css/browserOutput_override.css
+++ b/code/modules/goonchat/browserassets/css/browserOutput_override.css
@@ -62,6 +62,7 @@
 .marqua					{color: #00FFFF}
 .akula					{color: #f8412c}
 .plant					{color: #882d17}
+.opifex					{color: #aa00aa}
 
 .good                   {color: #4f8529; font-weight: bold;}
 .bad                    {color: #ee0000; font-weight: bold;}
--- a/code/modules/goonchat/browserassets/css/browserOutput_override.css
+++ b/code/modules/goonchat/browserassets/css/browserOutput_override.css
@@ -59,10 +59,10 @@
 .romana					{color: #ed7961}
 .euro					{color: #6533da}
 .yassari				{color: #5fbf4e}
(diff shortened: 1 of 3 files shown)


OliOliOnsiPree:
> Can you get some screen shots how this looks in a game on dark/light mode then next to science channel?

image
image
It might be a bit too similar to Janas color however, can always make it more deep if needed. The current color in this PR just takes from what Bay had for their bird species.

OliOliOnsiPree:
Color has changed to red on suggested by Trilby
image
image

Rcolor has been changed from Marqua
image
image

du-top:
Small question, not to be taken as criticism: Didn't people want Opifex to be more unique and LESS than Vox and now you change their language sounds to be exactly Vox? Am I missing something?

OliOliOnsiPree:
> Small question, not to be taken as criticism: Didn't people want Opifex to be more unique and LESS than Vox and now you change their language sounds to be exactly Vox? Am I missing something?

So to answer your question. So opifex language the old one that has been replaced in this PR, was taken from the cultist. Which Crayon Magic uses. Which copypasting other languages that are say used or having the same color per say. So example. If you talk in Opifexee and Cult speak now on the server. You can barely tell the difference unless you know one of the languages or the way they speak. Which sure sure, reverting them back to the language they "originally" had, is a step backwards, but its still original from what they have now. Which again is cult speak.

cdb-is-not-good:
this is fine.

PR #5097, opened 2024-03-05
Buffs the heavy pulse rifle

benj8560:
tweak making the heavy pulse rifle (the energy one guild makes via converting a rare exosuit pulse cannon) have the same damage output per cell as a Titanic. Essentially trades fire rate and shot capacity for per shot damage compared to the las-cannon finally making it a useful (if still not very practical) gun.


cdb-is-not-good:
yea thats fine

PR #5121, opened 2024-03-20
By Lamasmaster - Fixes Tenacity and reduces Will to Survive's cooldown

Ryan180602:
## About The Pull Request
<details>
<summary>
	Made by Lamas, changelog below. Fixes tenacity and reduces Will2Survive's cooldown. Poke him for shit
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
fix: Fix Tenacity
tweak: Reduce Will to Survive's cooldown by 5 minutes
:cl:


cdb-is-not-good:
Is this fully tested? Trilby suggested that the way it's coded may result in a doubling up of shock threshold.

Lamasmaster:
It's hard to get a proper idea of it on my own however if it does infact double the crit threshold that means the other perk does too and that needs to be changed then entirely.
In which case both need to be not hardcoded into the proc but rather holder.soft/hardcrit.threshold += 20
From the tests it mostly works as if you had around 20 toughness on top as pain can still kneecap you all the same it just makes you be able to walk off bit more regular damage to yourself without slowdown starting.

PR #5130, opened 2024-03-29
Squishy heads of staff

benj8560:
Unlocks race restrictions on all LC roles for Aulvae at the request of CDB.


cdb-is-not-good:
Earmarking this for merging in let's say....early April.

benj8560:
seems early April to me

cdb-is-not-good:
It sure is. This is good to merge.

PR #5142, opened 2024-04-02
"Modern" screen size

Moltovz:
## About The Pull Request
Widens the play screen by five tiles, because by the absolute our status window is TOO big by modern standards.
Does what it says on the tin, makes the game window wider. (21 pixels wide instead of 15), the height is still 15. Sound system was also touched in this PR in the most hasty way possible. increasing audiable sound range from 15 to 17, which makes sound dropoff with range a bit nicer on this screen size. Citadel has a bunch of sound ranges, but this works fine and they're nerds anyway so who cares.

As for why touching the sound system was necessary: the old sound system was tied to screen size, where a resolution of 15x15 would make the sound range 15. Trying to make an abnormal resolution size (like 21x15) would cause anything that makes sound to runtime and cry.

image
image

Feel free to @ me in discord with issues, or suggestions for UI element placement.
<hr>

## Changelog
:cl:
tweak: game window size has been increased by five tiles
tweak: sound range is now a static number
/:cl:


Moltovz:
> Never used Git before so bare with me but this PR is near on impossible to work with for folks like me using a laptop. First two images are with the "fit viewport", last is how I'd play on standard. So basically the PR is going to force it so if I want to read chat, it's going to be difficult for me to see what's going on in the actual game. If I want to see what's going on in the game, I'm going to be squishing chat down that in busy chat scenarios, I'll never be able to keep up even if I upped the text size more. 


Could you clarify? They all look fine to read.

Dimasw99:
@MsRandylicious 
MIght need a irl picture to show it properly

king5327:
The main issue I have is that a lot of our mob awakening code is written on the assumption that the view square is 15x15.

Also I'm diametrically opposed to any non-square resolutions.

Moltovz:
NEVERMIND Eris is doing this AND MORE and a REAL CODER is doing it!

PR #5158, opened 2024-04-08
LC personal shields tweak.

benj8560:
removal: Comments out the code related to the primes fancy personal shield icon. Its well agreed it very ugly. No one likes it. It gets in the way of everything visually. Now defaults to what all other personal shields use. Color adjusted to a church goldish yellow. Icon itself left alone and the codes still there if anyone wants to make something better please do.

Balance tweak: Ups the range of the Primes shield by 1 tile radius. The church cares about protecting their own as a whole. Not just the Prime. 
While lowering the CROs personal shield by 1 radius on the same kind of logic. SI cares more about its valuable personnel asset then everyone else lets be real. Pretty much just flipping the two.

Screenie of what dis looks like: 
dreamseeker_qi8MayOXlb

### DIFF
--- a/code/game/mecha/equipment/tools/shields.dm
+++ b/code/game/mecha/equipment/tools/shields.dm
@@ -98,7 +98,7 @@
 	desc = "A wide shield, which has the property to block incoming projectiles but allow outgoing projectiles to pass it.  \
 	Slower moving objects are not blocked, so people can walk in and out of the barrier, and things can be thrown into and out \
 	of it."
-	icon_state = "church_shield"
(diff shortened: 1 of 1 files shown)


benj8560:
forgetting the removal label :D

SingingSpock:
Ooh, I like the range buff. It definitely makes having a Prime on the field (or someone given that device) more of a general battlefield asset for the colony in a way that isn't just direct DPS.

benj8560:
> Ooh, I like the range buff. It definitely makes having a Prime on the field (or someone given that device) more of a general battlefield asset for the colony in a way that isn't just direct DPS.

Yes it's bothered me ever since it was added close range shield like that just doesn't make any sense for church. It's in reality actually a nerf to the prime themselves. Smaller shield is better in practice. But they can shield others now fits lore better.

PR #5180, opened 2024-04-14
reducing the odds of explosive trap spawns in maints and other areas

DustanHache:
## About The Pull Request
<details>
<summary> Explosive traps have in general left a bad taste in my mouth, and this can rightfully be considered a salt PR. My main gripe however is that they are simply too **common** for what is supposed to be a semi-secured colony with a dedicated military and security force. The goal of this is to shift the odds to less brutal traps to lessen the chance that someone’s first step into the maintenance tunnels becomes their immediate last and destroys a section of the nearby hall (since there are now trap spawns under basically every maintenance doorway and apparently the kitchen storage too.)
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
balance: Blackshield and Marshals, after numerous complaints from both medical personnel and colonists, have drafted up new IED and ordinance disposal procedures for their teams between shifts. Mines should be less common, though the trappers responsible have simply started placing more mechanical traps instead.
/:cl:

### DIFF
--- a/code/game/objects/random/traps/traps.dm
+++ b/code/game/objects/random/traps/traps.dm
@@ -5,10 +5,10 @@
 
 /obj/random/traps/item_to_spawn()
 	var/list/possible_traps = list(/obj/structure/wire_splicing = 1,
-	/obj/item/mine/armed = 0.15,
-	/obj/item/mine/improvised/armed = 0.30,
-	/obj/item/beartrap/armed = 0.45,
-	/obj/item/beartrap/makeshift/armed = 0.8)
+	/obj/item/mine/armed = 0.05,
+	/obj/item/mine/improvised/armed = 0.10,
+	/obj/item/beartrap/armed = 0.65,
+	/obj/item/beartrap/makeshift/armed = 0.9)
 
 	//Check that its possible to spawn the chosen trap at this location
 	while (possible_traps.len)


DustanHache:
Note that this is more of a stopgap offering until I learn the mapping tools, or someone else does a pass over the random trap spawns. It should NOT be fullmerged without really good reason.

Rat-Time:
This will change the spawn of random mines on every map, won't it?  Dungeons/forest?

DustanHache:
> This will change the spawn of random mines on every map, won't it? Dungeons/forest?
 Short answer, yes.
Long answer: That is personally something I don’t mind. Ideally this PR is closed after the colony based trap spawns are little more staggered, with explosives in particular not being able to spawn directly under doors leading to maints from the colony. halls

Rat-Time:
Could make a new entry for random traps specific to the colony and then swap that in for the colony map.

Rat-Time:
StrongDMM_zd6dHTuPwt
StrongDMM_WfdpR2xE12
These are the four that exist and have a 70% of not spawning anything.

DustanHache:
> Note that this is more of a stopgap offering until I learn the mapping tools, or someone else does a pass over the random trap spawns. It should NOT be fullmerged without really good reason.

This statement should clarify that yes, my intent is to remove/change those spawners, but there's definitely more than just those four. Additionally, there may or may not be bugs/stacked spawners in some locations, which cause multiple mines to spawn on the same tile. As soon as I figure out the mapping utility, I WILL be changing and/or closing this PR.

Rat-Time:
I'm stating that I have the map up and those four trap spawners are the only ones connected directly to a public area.  None are stacked and there are no trap spawners in the kitchen storage.  Deeper into floor two maintenance and all over floor three maintenance there are plenty of 100% spawn random traps but reducing the global probability of traps because of those four 30% chance traps doesn't seem right.  Additional traps you're seeing are likely placed by people mid-round.

DustanHache:
Not denying that as a possibility, though there are several other locations with shitty random trap spawns in that same area. Of particular note, ALL the doors north of RND in maintanance have a lowchance trap spawn, as well as another such spawn DIRECTLY AT THE BOTTOM OF THE STAIRWELL THERE.

cdb-is-not-good:
Not a huge fan tbh. These spawns are in maints, a disused area that's slightly dangerous but not deadly, the presence of random traps teaches people to not rush through doors

Trilbyspaceclone:
Does not fix the issue addressed and generally affects way more then perserves.

PR #5206, opened 2024-04-22
[Balancing] Lowers cooldown of Aulvaes limb regen

IrkallaEpsilon:
## About The Pull Request
A simple PR of lowering the Aulvaes limb regen from the insane 30 minutes to 5. This ability is not spamable anyhow due to needing 300 nutrients per use and it also is not a full heal. Also CDB kindly asked for a buff on slime limb regen in terms of a cooldown reduction.

## Changelog
:cl:
balance: Aulvae Limb Regen from 30 minutes down to 5. 
/:cl:


cdb-is-not-good:
450 nutri, five minutes....yea that's fine I guess. If it ends up being really abusable we will maybe wanna tone it back to like ten minutes but this should....be fine. Slimes should still be relatively fragile.

PR #5208, opened 2024-04-22
Sci shuttle Access

LongSleeves:
Gives soteria access to the surface transport shuttle.
Untitled


LongSleeves:
This was a PR that was requested by Knifetail (discord name)
I'm not actually sure who the shuttle belongs to in lore now cause i've seen a bunch of back and forth about it.
I just go on foot most days regardless.

cdb-is-not-good:
Oh yea I said this was probably fine tbh. Plus one

PR #5222, opened 2024-04-24
Re-adding Umi's Snow Tiles

YuiJaguar:
These Tiles are from LL that I never got to see in person because I left before launch. I want to see them used because they were very well received, so I've migrated them here for our mappers.

I'm very new to this so please be nice if I screwed up something. I have never really coded before. ~Umi

The icons themselves are decent quality when mixed together and can be rotated to create various patterns. 
Screenshot 2024-04-24 122041

They're muted enough to be used alone as well, for small maps.

Using Modular files together should create "pig trails" or walking trails like people do in the winter to get from spot to spot. 
Snow glades look nice if stacked 1-on 
These can *also* be used with our default snow icon I THINK but they are really meant to be used by themselves when looking at snow.


YuiJaguar:
Ah and, yes, I didn't make it clear but I am the original artist for all these tiles. I didn't just rip it from LL. These were submitted for use by me to LL back when it was in it's compilation stages, and accepted. I however left the server before I got to see them being used. So these are my original tiles.

PR #5246, opened 2024-04-29
Xenopit Steward Headset and Misc Brig Changes

Lamasmaster:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

Moves the dream daemon headset to the xeno pit, alongside adding spawn of armor plating and low chance for another gun, reasoning being that at least this way you fight for it instead of killing 1 DD and sometimes not even that after mining in for all comms access.
image

Recreation room for security is now it's own area so you don't need to wonder why you need to go down to blackshield tactical armory to turn power back on. Also made more easier to traverse at loss of few vendors.
image

Supply specialist now spawns outside the armory/armory shop because how did he even get in here when the shutters are all closed, redesigned it a bit to make it more traversable, ablative / bulletproof armors and thermal helmets are in same place now inside of boxes so they can be handed out easier. APC is moved outside of emergency armory and into the workshop.
image

TACARM gear reduced to 4 of each of the batons/armors everything else same count, hopefully this will make it way more traversable and neater to actually look at.
image

<hr>
</details>

## Changelog
:cl:
tweak: Xenopit / dream daemon den, brig
/:cl:


cdb-is-not-good:
Yea this is fine. Just quickmerge so it doesn't sit there blocking other mapping PRs

PR #5247, opened 2024-04-29
TGUI Crafting Menu

ShadowLarkens:
## About The Pull Request
<details>
<summary>
Converts the personal crafting menu to TGUI. Based on /tg/'s interface, but made custom.
</summary>
<hr>

| Before | After |
|-------|-------|
|Before|After|

Includes searching support.
Searching Support

And a compact mode.
Compact Mode

	
<hr>
</details>

## Changelog
:cl:
add: Crafting menu now uses TGUI

/:cl:


ShadowLarkens:
As a sidenote, I've introduced a tgui_module class here, based on nano_module, from Virgo, but it definitely isn't necessary for this particular interface. I think later, I might convert this interface to work the same way it does on /tg/, which is to say make it a /datum/component that is automatically added to humans and robots.

PR #5248, opened 2024-04-30
EMP scaling

Lamasmaster:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

Ports the organ damage function code from this PR
https://github.com/Baystation12/Baystation12/pull/34346
This makes FBPs / Anyone with prosthetic organs take less direct damage from the EMP itself making it more into a disable due to all the wounds it still causes rather than being instant kill.

<hr>
</details>

## Changelog
:cl:
balance: ION/EMP damage
/:cl:

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -292,16 +292,17 @@
 		module.activate(owner, src)
 
 /obj/item/organ/external/emp_act(severity)
+	var/rand_modifier = rand(1,3)
 	if(!BP_IS_ROBOTIC(src))
 		return
 
 	switch (severity)
 		if (1)
-			take_damage(20, BURN)
+			take_damage(8 * rand_modifier, BURN)
 		if (2)
-			take_damage(10, BURN)
+			take_damage(6 * rand_modifier, BURN)
 		if (3)
-			take_damage(5, BURN)
+			take_damage(4 * rand_modifier, BURN)
 
 /obj/item/organ/external/attack_self(var/mob/user)
 	if(!contents.len)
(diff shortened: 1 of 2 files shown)


cdb-is-not-good:
>Makes EMPs on average way instakill win button to FBPs while still maintaining their strong disabling power

Reword this please, it makes no sense what you mean

Dimasw99:
This sadly wouldn't balance or fix EMP's against robotics by any significant amount since the major issue is that EMP's affect every single internal organ at once.

What I would do instead is make internal robotic organs only take damage if the exterior robotic organ has a set amount of burn damage.
So if I have a fleshy arm with a synthetic muscle, it gets fried by EMP
If the arm is robotic and gets hit by an emp, it first gets absorbed by the arm and then after x amount of damage, applies to internals.

Lamasmaster:
What it would do is put up a stopgap between even least severe EMPs instantly breaking certain organs like nerves while a proper rework of ions can be made, not saying it's perfect but it's something at least. As is right now getting sneezed at by EMP some of which still got huge radius like rockets means death which isn't fun for the recieving end if I had to imagine.

cdb-is-not-good:
It's probably an acceptable first step.

Dimasw99:
> It's probably an acceptable first step.

True, even in a super scuffed way.

Locking through EMP code it is pretty funny. Most EMP's use either severity 1 or 2 with only a single instance using severity 3 for APC's specfic alone. The math for wounds similarly is just the current damage amount divided by 4. Making the solution a somewhat pure pflaster fix on a giant gashing wound. The end result is that in severity 1 cases, you have 33% to roll 0 wounds, 33% to roll 1 and remainder to roll 2 wounds. With severity 2 it is a case of 66% of not rolling a wound. Sooo in the end, an FBP will still have one third of all its organs die in the best case scenario and with the current wound system it is very unforgiving

cdb-is-not-good:
I see no reason to not fullmerge this for now. We can further pare back EMPs and tweak them later.

PR #5250, opened 2024-04-30
SLT sidearm options

Lamasmaster:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

Gives SLTs the option to choose between the Abnegate, soteria version of the Zeus and Spider Rose.
Spider Rose doesn't see much use neither does Zeus with Zeus being an orderly weapon at one point so it makes sense.
	
<hr>
</details>

## Changelog
:cl:
balance: Sidearm options for SLTs
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/hardcases.dm
+++ b/code/game/objects/items/weapons/storage/hardcases.dm
@@ -442,11 +442,10 @@ obj/item/storage/hcases/attackby(obj/item/W, mob/user)
 
 /obj/item/storage/hcases/med/medical_job_trama/populate_contents()
 	new /obj/item/gearbox/traumatizedteam(src)
+	new /obj/item/gunbox/traumatizedteam_sidearm(src)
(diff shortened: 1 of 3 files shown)


cdb-is-not-good:
Not that I really care much but why would one ever pick Zeus over the spider rose which has hitscan and a lethal mode

SingingSpock:
With its current cell, the Hera will get 5 shots per cell. With a posi, it gets 8.

cdb-is-not-good:
Hum, well im not a huge fan of giving them a sidearm that *is* potentially lethal. They don't need that kind of versatility.

Lamasmaster:
I could just remove the spider doesn't really matter just seemed like it would fit, Zeus has enough shots to down a single person reliably as long as they aren't on oxy.

Lamasmaster:
Removed spider

PR #5251, opened 2024-04-30
spin fix

Dimasw99:
## About The Pull Request
Port spin fix from https://github.com/Baystation12/Baystation12/pull/33885 that would otherwise cause a sleep(0) to loop endlessly if called by just spin()
Kudos to DRD for causing another crash and shame on chefdoggo for being the fool to call it (jk love you)


ChefDoggo:
server
explodes server with advanced proc call

PR #5306, opened 2024-05-18
marshal break room fix

Dimasw99:
## About The Pull Request
Someone copy pasted a maints APC into the marshals break room, making it off by default.
Also remove a random spider spawner inside the walls.
grafik


Trilbyspaceclone:
NOOO! NOT THE WALL SPIDERS!
That has survived for like 5 years! ITS A RELIC!
It was added by the O.G map maker day one :~: keep the history!

Dimasw99:
> NOOO! NOT THE WALL SPIDERS! That has survived for like 5 years! ITS A RELIC! It was added by the O.G map maker day one :~: keep the history!

grafik
turned it into a funni thing then

PR #5314, opened 2024-05-19
Vexilar Forceblade Rework

Xaytan:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>
Reworks the Vexilar Forceblade. It now uses 5 power from the user's cruciform to activate, and then an additional 5 power with every attack. It should now more suitably serve its purpose as an endgame Church weapon.

Now projects a glow when turned on, as well as being hot for the purposes of fuel spills and plasma.

Added a 20% damage increase when used against hive creatures and psi monsters. Deus Vult!

	
<hr>
</details>

## Changelog
:cl:
balance: Reworked Vexilar forceblade power consumption to be per-hit.
balance: Added Vexilar Forceblade damage increase when attacking Hivemind or Psi Monsters
balance: Increased material cost of Vexilar Forceblade.
balance: Added glow effect while Vexilar Forceblade is toggled on.
:cl:


cdb-is-not-good:
neat, big plus one if it's actually tested and working.

PR #5327, opened 2024-05-22
rework of damage mods

Dimasw99:
## About The Pull Request
Reworked how damage modifiers are applied to people, making scaling linear instead of being a 1/x curve.
All damage modifiers are now multiplicative instead of additive.

Reasoning: For a while now I noticed that a lot of damage modifiers are tip toeing the line between incredibly small but if stacked enough will utterly break the balance. For example if you have a brute mod of 0.2 from prior stacking, adding another -0.1 from something like cow skin would double your effective HP. Otherwise, it only would be 10% if you had nothing else.
Now that they are multpilied, it is much easier to see how much damage you take less while also preventing the insane stacking. Another reason was that a lot of choices for damage absorbtion by itself were more than just meaningless, the game was encouraging to stack and power game to make any good use out of it.
grafik
Of course I did the basic math behind it, coloumn B is our current additive system. You can push it to recieving only 12% but it gets worse since you can exploit resurrection sickness and gigantism to gelow below it. Likewise you can go below it by using the churches abillity.
Coloumn C is the basic unadjusted version of B, 1 - old number
Coloumn D is the manually adjusted rebalance and rebuff of each option since we no longer have to fear insane scaling later on, showing that even with adjusted numbers this is technically a nerf.

### DIFF
--- a/code/datums/perks/job.dm
+++ b/code/datums/perks/job.dm
@@ -284,20 +284,20 @@
 	..()
 	initial_time = world.time
 	cooldown_time = world.time + 30 MINUTES
-	holder.brute_mod_perk += 0.10
-	holder.burn_mod_perk += 0.10
-	holder.oxy_mod_perk += 0.10
-	holder.toxin_mod_perk += 0.10
+	holder.brute_mod_perk *= 1.10
+	holder.burn_mod_perk *= 1.10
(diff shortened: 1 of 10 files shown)


hex37:
I'm fine with the changes, WRT genetics.

PR #5332, opened 2024-05-23
Underplating tripping checks higher of vig or cog

Xaytan:
## About The Pull Request
<details>
<summary>
	About The Pull Request

</summary>
<hr>
Encourages build diversity by allowing another stat as a potential trip bypass, allowing more characters to take interesting backgrounds that aren't Former Scavenger.

The higher of vigilance or cognition is used to determine trip chance. Dumb melee builds can WEEP.

Gives an additional reason to not dump vigilance, since doing so is kindof all the rage right now.

	
<hr>
</details>

## Changelog
:cl:
balance: Vigilance allows the user to avoid tripping on underplating.
/:cl:

### DIFF
--- a/code/game/turfs/flooring/flooring.dm
+++ b/code/game/turfs/flooring/flooring.dm
@@ -278,7 +278,8 @@ var/list/flooring_types
 		if(M.stats.getPerk(PERK_SURE_STEP))
 			return
  // The art of calculating the vectors required to avoid tripping on the metal beams requires big quantities of brain power
-		if(prob(50 - our_trippah.stats.getStat(STAT_COG))) //50 cog makes you unable to trip
+ // The truly aware can smell the rolled ankle's unwashed bald spot and dodge it like they dodged the draft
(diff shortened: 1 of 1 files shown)


Xaytan:
Tested locally, works fine on my machine...

cdb-is-not-good:
Vig is....already way too strong. I don't really like this change. Id rather you just flat reduced the chance of tripping on under plating across the board.

Xaytan:
> Vig is....already way too strong. I don't really like this change. Id rather you just flat reduced the chance of tripping on under plating across the board.

We could. Lowering the cog requirement to avoid tripping to 30, maybe?

Currently the meta powergaming strat is to dump vig as low as possible if you can get away with it, or to take no higher than 40 vig (Enough to wield most weapons instantly) if you're playing a role like Blackshield that needs that.

Could also make the calculation take the average of vig and cog if we really wanted to be evil...

hex37:
Personally against this change- cog has so little it's useful for, let it have one of the few things its good for!

cdb-is-not-good:
> Then make Cog finally affect sanity and fear related things instead of vigilance then so that way it's at least consistent with being the brain stat instead of not making the agility stat affect agility related feats

That's probably a better move, yes.

PR #5339, opened 2024-05-24
[bugfix] NO_PAIN flag now actually properly confers pain immunity.

cdb-is-not-good:
Previously, the nopain flag/mutations didn't actually stop you from accumulating halloss(which meant you still suffered all kinds of deleterious effects). This has been fixed. This will also introduce a new issue because some jackalope decided to make it so that shields cause halloss when struck even though halloss is used as pain(not STAMINA). this i will maybe fix later.

## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -203,6 +203,11 @@
 /mob/living/carbon/human/setToxLoss()
 	return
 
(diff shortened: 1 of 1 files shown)


Dimasw99:
I think halloss is sort of like stamina already or just simple bodily exhaustion. We don't have another var for that from what I know

PR #5340, opened 2024-05-24
Rebalance recluse to be 2 hit instant death

Dimasw99:
## About The Pull Request
Certified salt PR moment. Why do we have an instant death spider with no way to proper counter play? Prob Kaz thought it was funny. Good thing he is no longer around!

Now you need at least two hits to be dead and zombie powder as well requires 1 unit to work, preventing players to cheese super thinned out zombie powder to sleep people indefinetly. So if you do end up lagging into them (Some of us play with a 1 second stutter like me :( ) you don't end up instant dead.

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/giant_spider/types/nurse.dm
+++ b/code/modules/mob/living/carbon/superior_animal/giant_spider/types/nurse.dm
@@ -97,7 +97,7 @@
 	icon_living = "recluse"
 	maxHealth = 15
 	health = 15
-	poison_per_bite = 0.5 //Lets not have a classic ss13 traitor chem be used as an instant down when people are lagging. 2 hits for your unfun spider
+	poison_per_bite = 1
 	melee_damage_lower = 3
 	melee_damage_upper = 5
 	meat_type = /obj/item/reagent_containers/food/snacks/meat/spider/recluse
--- a/code/modules/mob/living/carbon/superior_animal/giant_spider/types/nurse.dm
+++ b/code/modules/mob/living/carbon/superior_animal/giant_spider/types/nurse.dm
@@ -97,7 +97,7 @@
 	icon_living = "recluse"
 	maxHealth = 15
 	health = 15
-	poison_per_bite = 1 //1u is all it takes to sleep you, your asleep also dosnt prevet it form attacking you more then once meaning this quit quickly stacks without someones help
+	poison_per_bite = 0.5 //Lets not have a classic ss13 traitor chem be used as an instant down when people are lagging. 2 hits for your unfun spider
 	melee_damage_lower = 3
 	melee_damage_upper = 5
 	meat_type = /obj/item/reagent_containers/food/snacks/meat/spider/recluse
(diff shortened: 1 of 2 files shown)


Dimasw99:
> Hrm. I'm kinda iffy on this. Did you test this PR while using armor? At what level of armor are you functionally immune. Remember, armor should reduce the amount of chem injected

Your fancy CO armor. Though that is the hilarious part, **ANY** amount of zombie powder will instantly knock you out. With a MET of 0.2 the lowest amount needed to stun you is just 0.02 units of it in your blood.

Dimasw99:
> Hrm. I'm kinda iffy on this. Did you test this PR while using armor? At what level of armor are you functionally immune. Remember, armor should reduce the amount of chem injected

grafik
You need 99 armor to block the toxin lmao, so even with 98 armor you would get insta stunned since 0.02 is still enough for a single cycle. Which then applies a weaken 10 onto you :p

cdb-is-not-good:
> > Hrm. I'm kinda iffy on this. Did you test this PR while using armor? At what level of armor are you functionally immune. Remember, armor should reduce the amount of chem injected
> 
> grafik
> You need 99 armor to block the toxin lmao, so even with 98 armor you would get insta stunned since 0.02 is still enough for a single cycle. Which then applies a weaken 10 onto you :p

No no I mean how much will work WITH the changes

cdb-is-not-good:
Like with your changes how many bites will it take for say ...standard armor vest, or riot gear, etc. I want to know how this change will play for someone who isnt defenseless

Dimasw99:
Riot armor is a designed anti melee protection and not that easy to get with 60 melee protection and being able to be upgraded to 80. So the math would simply be 1/(0.5 * 40%) = 5 hits. Flak armor with 40 melee would take 4 hits. 20 melee protection which is usually your super cheap armor types require 3 hits to die.

So going back to upgraded riot armor or even ameridian rig armor, it takes 10 hits to die. Which in all fairness, is justified considering one is a specifically ment to counter that type and the other is an over complicated and expensive armor

cdb-is-not-good:
So your PR title is a bit misleading. I'm not against making zombie powder not proc on any amount, but if you're gonna do that at the very least make it so that having basic armor is what makes it take two bites not having none(and thus meaning any serious amount of armor trivializes this enemy.)

cdb-is-not-good:
Especially seeing as, y'know. Recluses are made of tissue paper.

cdb-is-not-good:
...also a standard armor vest has 30

code/modules/mob/living/carbon/superior_animal/giant_spider/types/nurse.dm
```diff
@@ -97,7 +97,7 @@
 	icon_living = "recluse"
 	maxHealth = 15
 	health = 15
-	poison_per_bite = 1 //1u is all it takes to sleep you, your asleep also dosnt prevet it form attacking you more then once meaning this quit quickly stacks without someones help
+	poison_per_bite = 0.5 //Lets not have a classic ss13 traitor chem be used as an instant down when people are lagging. 2 hits for your unfun spider
```
cdb-is-not-good, line 100:
```suggestion
	poison_per_bite = 1
```

Seeing as how we have the armor reduces injection code, both reducing the dose *and* adding the code that makes doses lower than 1 not take effect, it isn't necessary to do both. If you go into maints without armor, that's an unforced error.

cdb-is-not-good:
Mkay, im satisfied. For now, please wait as i need to get some testing done for the combat PR(and that includes fixing iinjection code vs armor)

PR #5344, opened 2024-05-25
Projectiles go over disposal bins

WatermelonsEverywhere:
Projectiles now fly over disposal bins.

----------------

I'm not completely happy with just adding a flat projectile check, but it seems that we don't use the CanPass() "height" variable anywhere in the code for these sorts of purposes, and I don't want to do a code crawl and adjust every single copy-pasted "height=0" in the code.

### DIFF
--- a/code/modules/recycling/disposal.dm
+++ b/code/modules/recycling/disposal.dm
@@ -467,6 +467,9 @@
 				visible_message("\The [I] lands in \the [src].")
 			else
 				visible_message("\The [I] bounces off of \the [src]\'s rim!")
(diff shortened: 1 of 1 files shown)


cdb-is-not-good:
Well, if it works it works.

Trilbyspaceclone:
image
Was this tested?

WatermelonsEverywhere:
Yes, bullets and lasers went straight over, I imagine they aren't internally flagged as "currently being thrown", so I'm unsure what that above check even is for. I couldn't find a case of a projectile entity being thrown, but I decided not to touch it just in case.

PR #5345, opened 2024-05-25
Xenoflora-Xenobiology QOL

TheEternalFlame:
## About The Pull Request
<details>

 Xenobiology had a corestopper added at round start and adjusts the charge cost from 100 to 20 (1/5th)

Xenoflora got a full makeover making it generally better for use, farmbots shouldn't get stuck anymore, while making the room more open by removing the animal room from the back that was there for some reason.

<summary>
Merges the two PRs touching xenoflora and xenobiology.
</summary>
<hr>

Xenobiology has had a slime compressor to the surgery area, and a corestopper placed on the table in the middle of all the slimes and the charge cost to fire them brought down from 100 to 20, meaning that instead of the previous 4 shots you would get with a 400s small cell you'll now get 20 shots before needing to recharge it. Given that the corestopper is only useful against slimes (Not the people ones.) there was no point in keeping the amount of shots that low, especially when breeding slimes they can get out of hand so quickly that you'd virtually never be able to keep up with them using a corestopper unless you had 10-15 spare cells just on hand to switch out.

Xenoflora has had a full facelift, while keeping the same amount of farm trays, the atmos secure room has been moved off to the west side of the room and placed behind a non-glass door and walls so that farmbots won't get confused and try to reach them while being unable to. The eight trays that were in the front of the room have been moved to the back and spread out with a 1 tile gap between them all.  The animals at the back have been removed along with their enclosure and the chemistry machines, plant modders and vendors placed along the north wall to use the room left by it. An advanced soteria grinder has also been added to help with the grinding and producing of bottles in a more clean and clear way, given the garden has an industrial grinder it only made sense for Soteria to use their own grinder for their xenoflora bay.

image


du-top:
Still no fire extinguishers or oxygen cabinets?

Trilbyspaceclone:
image
The o'ed lights could be unshifted

Trilbyspaceclone:
image

du-top:
> And there's a fire extinguisher right beside the vendors on the wall.
Goddamn. There is. Even with floor marking. Surprising that I didn't die in a fire yet.

Dimasw99:
The APC is starting depowered. You may have copied a maintenance APC

Dimasw99:
grafik
Could you make these two tiles transparent as well? One windowed door and the other just a simple window

Trilbyspaceclone:
ima rework the current verson and we can have a vote on what ones better if need be

du-top:
> Could you make these two tiles transparent as well?

This is on purpose, the farmbot is not supposed to see the trays inside.

Trilbyspaceclone:
image
As per the vote A has one

PR #5350, opened 2024-05-25
NSA icon is now percentage based

Dimasw99:
## About The Pull Request
NSA UI now respects modified thresholds like buffs and debuffs to it


cdb-is-not-good:
tested and works?

PR #5374, opened 2024-05-29
Removes the spare Church vendor

SingingSpock:
## About The Pull Request
<details>
<summary>
	Removes the Church vendor from upstairs
</summary>
<hr>

It's just used for free disks and such rather than people using the EOTP which is the intent. I've also heard people are recycling things from it. It doesn't really serve a needed purpose, so it can go.

image
	
<hr>
</details>

## Changelog
:cl:
map: Removes the upstairs church vendor that Churchies only use for free disks and such
/:cl:

### DIFF
--- a/maps/__Nadezhda/map/_Nadezhda_Colony_New.dmm
+++ b/maps/__Nadezhda/map/_Nadezhda_Colony_New.dmm
@@ -94727,10 +94727,10 @@
 /turf/simulated/floor/tiled/steel/techfloor,
 /area/nadezhda/engineering/workshop)
 "tyH" = (
-/obj/machinery/vending/theomat,
 /obj/structure/sign/faction/neotheology_cross/gold{
 	pixel_x = 32
 	},
+/obj/structure/table/standard,
 /turf/simulated/floor/tiled/dark/techfloor_grid,
 /area/nadezhda/absolutism/hallways)
 "tyI" = (
--- a/maps/__Nadezhda/map/_Nadezhda_Colony_New.dmm
+++ b/maps/__Nadezhda/map/_Nadezhda_Colony_New.dmm
@@ -53818,6 +53818,7 @@
 	icon_state = "1-8"
 	},
 /obj/effect/decal/cleanable/dirt,
+/obj/structure/table/rack,
 /turf/simulated/floor/carpet/bcarpet,
 /area/nadezhda/absolutism/bioreactor)
 "lij" = (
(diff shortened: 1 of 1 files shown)


benj8560:
forgot a wont fix label. Theres another vender a short walk into maint they can just grab and likely will. However. I think just removing the mechanic of them being able to unlock it is a better fix. The mechanic itself was ported from Eris. A place where that vender contains oh boy. 20 wine bottles and some books crazy stuff.

Either route I don't think we have any bottles of Cahors outside of those venders in the church? Vectors should not really need to pay for just the bottles and something most make a EDC.

SingingSpock:
> forgot a wont fix label. Theres another vender a short walk into maint they can just grab and likely will. However. I think just removing the mechanic of them being able to unlock it is a better fix. The mechanic itself was ported from Eris. A place where that vender contains oh boy. 20 wine bottles and some books crazy stuff.
> 
> Either route I don't think we have any bottles of Cahors outside of those venders in the church? Vectors should not really need to pay for just the bottles and something most make a EDC.

The vendors being able to be unlocked, their prices set and more items added is good functionality for all of these vendors, and I don't really want to remove it from everyone to stop Church from doing one minor thing. If they start regularly grabbing the maint one for the purposes of taking free stuff from it, that one can be removed too without much loss to the colony.

SingingSpock:
image
One of the barrels has been replaced with 10 bottles

benj8560:
as I said. Nothing changed. They already just started grabbing the vender in maints. :P

benj8560:
the unlocking the vender to change prices and such is nice yes. Problem is. Its not used for that **ever** church makes a new vender when they wanna do that.

benj8560:
then you remove that one. I assure you the next step will be they just walk out to the chapel and grab the disks from the public vender. Nothing says they can't.

SingingSpock:
> the unlocking the vender to change prices and such is nice yes. Problem is. Its not used for that **ever** church makes a new vender when they wanna do that.

But this isn't a feature unique to Church vendors, it's universal. Making the Church the only vendor that doesn't do this would be weird.

benj8560:
> But this isn't a feature unique to Church vendors, it's universal. Making the Church the only vendor that doesn't do this would be weird.

So I should give guild access to their vender so they have round start mark Vs?

benj8560:
or what about the marshal vender so specs can just recycle plasteel from there to feed their armory?

SingingSpock:
Ah. See, I was under the impression that's how it worked for all faction vendors. I'll just remove this functionality then, it's bad.

PR #5392, opened 2024-06-01
Adds area to SLT helmet alerts, suit sensor changes

WatermelonsEverywhere:
The special SLT alert helmets not also print the area the alert came from.

Suit sensors can now be changed with tools that have "SCREWDRIVING" but are also sharp, like knives.

Only clothing that spawns on Station Levels (The colony, the mines, the solars) start with sensors enabled, clothing that spawns outside starts with sensors disabled.


Trilbyspaceclone:
With your changes can you still cut suits up with knifes?

WatermelonsEverywhere:
I don't believe this has an effect on that since it's a crafting recipe,
and starting it turns the jumpsuit into a crafting pseudo-item that
shouldn't be affected. Is there a non-crafting interaction I'm not aware of?

PR #5412, opened 2024-06-05
Swaps Fence Bison for a Vector SMG and gives the foreman ball 12mm

benj8560:
mini balance PR for Fence. I always strongly disliked the Fence getting a uncommon and exotic guild craft weapon. After some back and forth me, are currently only regular fence player Decker and CDB agreed on a hard spawn Vector SMG with 2 ball and 1 hollowpoint SMG mag. Should be still plenty deadly as long as the new combat test merge sticks around. Also gives the foreman ball rather then 12mm a friend asked. HV is really really real bad compared to ball and in their line of work now.

Changelog:

Replaces the fences Bison shotgun with a Vector SMG.
Replaces foremans HV ammo with ball cos HV is bad now


cdb-is-not-good:
folded into the bulk combat PR.

benj8560:
due to heavily depending on it for intended balance if anyones wonderin

PR #5414, opened 2024-06-06
Nanite Rigsuit can now be removed, also refactored

ShadowLarkens:
## About The Pull Request
The nanite rigsuit has had a bug where you can remove it after toggling it once, because rigsuits use `canremove` in a really cursed way. Rather than try to fix this, this just allows you to always take it on and off freely. 

Additionally, the nanite rigsuit was doing some really cursed overrides and copy-pasting code from base rigsuit procs, I refactored it to not do that, and instead just override a new `can_maintenance()` proc.

## Changelog
:cl:
tweak: Nanite Rigsuit can now be freely removed, rather than buggy sometimes removal
refactor: Nanite Rigsuit no longer uses a bunch of copy-pasted code
/:cl:


cdb-is-not-good:
Hrm, I'm not sure I like changing the rigsuit being removable just because it's easier tbh. But it's also kinda whatever.

PR #5419, opened 2024-06-07
Fixes SI outpost atmos setup

benj8560:
Was curious just what exactly was sucking so much power at SI outpost. Turns out! toxins was not all to blame. It's atmos was setup in such a way that it attempted to compress via filtering several very very large air canisters into a single pipe. Making them run non-stop all round.

This PR fixes that! Some things shuffled around. Nothing removed/added unrelated to atmos. Did testing via dumping every kind of gas into the system and making sure it went to the right places.

Changelog:

Mildly shuffles around and fixes SI outpost atmos making it no longer consume a absurd amount of power and now actually functions.

Pretty picture: 
StrongDMM_2My4wZHoSg


benj8560:
and in case anyone's curious on examples of what not to do all the air from all 4 large air tanks was being compressed as show here. 
ShareX_phe5uWiJfD

benj8560:
....poorly shown now that i look at it again. To be clear they filters where set to output east.

PR #5429, opened 2024-06-09
Adds Paradise's Click Catcher & Misc. Fixes

### LINKED ISSUE #5416: Door + remote signaler issue
Doors don't accept remote signalers anymore when hacking (attach button), does not matter if the signaler is set to attach or not, since this is only used to attach to other things like prox sensors
image
image

ShadowLarkens:
## About The Pull Request
clicking in darkness

This adds Paradise's implementation of the click catcher to allow you to fire and click where your vision is obscured. This version doesn't rely on Eris's god awful system that creates 225 individual screen objects - instead, it just stretches a big ol' sprite across the whole screen and puts it behind everything else.

Also fixes:
- Signalers being unable to attach to wires. (Fixes #5416)
- Radio headsets making noise even when you can't actually hear the channel (out of TComms range)

## Changelog
:cl:
fix: Signalers can now be attached to wires again
fix: Radio headsets no longer make the radio noise for messages you can't actually receive
balance: Click Catcher re-added, with a better implementation that should not lag and should provide good clickies in darkness
/:cl:


Dissentbay:
#5416 is gone! 🙏

PR #5433, opened 2024-06-10
Rig suit QoL

Dimasw99:
Rig boots have a slot for knifes.
Rig armor pieces can be deployed without power with a delay. 
Cells and tanks can be removed while wearing the rig. 
Rig armor pieces can be deployed over chest and head clothing if they don't possess significant armor. 
Fixed the bug where the rig suit would eat your clothing if it runs out of power. 
Rig gloves siemens coefficent is now multiplied by the gloves worn underneath.


cdb-is-not-good:
Love everything except swapping cells without removing the suit. This means that a major way of disabling rigs(EMP) is less impactful. Could this be- not removed, but tweaked to be slower or something? Just so you can't do it fast style

cdb-is-not-good:
absolutely based. if you can thoroughly test everything I'll fullmerge this tonight or so and fix the conflicts on my end.

Dimasw99:
Fully tested. Every single feature and change had its quirks ironed out (There were a few) and I even tested the potential meme of having 2 backpacks which is also luckily not possible.

PR #5439, opened 2024-06-10
Trash™ update

GlitchySoftware:
After an hour of figuring out how to revert things, half-assedly I got it back and here we are, adds some greenery and overgrowth to more outposts, also renders made their way into the burned down outpost.. bringing a bit more loot with them.
image
image
image


TheEternalFlame:
For the love of god **PLEASE** do not put renders near the trainee claim.

GlitchySoftware:
> For the love of god **PLEASE** do not put renders near the trainee claim.

:3 spiders will suffice then, a lot of spiders..

Dimasw99:
uuuuh 
grafik

PR #5443, opened 2024-06-11
[Balance I guess?] Changes the defunct toxloss component of 2 gun mod to Halloss

IrkallaEpsilon:
Both the Toxin Coater and the Glass Widows toxin damage components are defunct thanks to erismed IV. At least when it comes to pvp. 

## About The Pull Request
One could make an argument to just change it to apply toxic buildup like erismed IV does with roach gas for instance but Erismed IVs wound system isnt that well fleshed out to be honest as it increases in severity no matter how little a wound is. Flat damage numbers are far easier to balance around for gunmods. I also have grown to dislike (partially) broken equipment that is still available for printing and thus wasting materials or time. Glass widow remains 9 mm locked. 

Toxin coater has to compete with better barrel mods like the forged or heavy barrel.

Confirmed on local that halloss applies correctly.

To do in a separate PR: Figure out how to apply L.apply.effects onto guns because that would allow for a bunch of nice shit to work like irradiate (the proper rad barrel in code is not working and I think never really worked to begin with even upstream) or confuse. 

## Changelog
:cl:
tweak: Toxin coater and Glass Widow now apply halloss instead of toxin damage (Toxin damage is defunct, nope its not fixed upstream either :/)
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/mods/mods.dm
+++ b/code/modules/projectiles/guns/mods/mods.dm
@@ -296,7 +296,7 @@
 	var/datum/component/item_upgrade/I = AddComponent(/datum/component/item_upgrade)
 	I.weapon_upgrades = list(
 		GUN_UPGRADE_DAMAGE_BURN = 10,
-		GUN_UPGRADE_DAMAGE_HALLOSS = 10, //Once again this component was tox before, Tox damage does not exist in erismedIV and wounds are not particularly well thought out as we have seen with unmaker.
+		GUN_UPGRADE_DAMAGE_HALLOSS = 10, // Once again this component was tox before, Tox damage does not exist in erismedIV and wounds are not particularly well thought out as we have seen with unmaker.
 		UPGRADE_BULK = 1
 		)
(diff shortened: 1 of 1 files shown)


cdb-is-not-good:
Not really a fan, this feels like a bandaid fix to a greater problem of the question of toxin mod on weapons. Wouldn't it be more sensible to just make it so that it injects an amount of toxin that carries the toxicity level of the mod itself? It'd certainly be more work but if the fix is "make it do halloss and give up on making it work as intensed" then why even have it?

cdb-is-not-good:
What do you mean broken? Broken in what way?

cdb-is-not-good:
also can we not simply copy the on hit inject code from sopo/pepperball, or does that require making mods apply effects that they currently don't support

cdb-is-not-good:
Huh???...we *know* irradiate is working on the EBR

cdb-is-not-good:
Hm. Well first things first I'd say take a peek at Eris code and see if they still use this noise and if so what they did to make it compatible with their modern medical(if they even did). Copying their homework is fine

PR #5484, opened 2024-06-25
missed a few HEV things. Woops!

benj8560:
Gives the HEV suit cold protection and unacidable (for good measure). Should cover every environmental hazard other then space now.


Dimasw99:
Why no space proof?

benj8560:
> Why no space proof?

 Freeman never went to space

PR #5518, opened 2024-07-06
Handmade Guild Cells

Vex8tion:
## Changelog
:cl:
rcsadd: A line of 'Workhorse' hand-made Guild power cells.
/:cl:

Adds a line of handmade cells for the Guild. These come in a 300S, 1200M, and 14000L version. They're crafted with gold, but charge twice as fast as standard Robustcells or Power-Geysers as a result.


hex37:
Sorry to say- I'm definitely not a fan of this PR- Batteries are Soteria's thing and should stay that way.

HanSolo1519:
Not a definitive no by any means, however, if it is decided that these clash with Soteria's battery-niche, I recommend making them craftable using high-end soteria cells, similar to how the handforged parts need interaction with soteria for tier 3 parts, these could require power-geysers.

I do love the sprites, they do fit in very well

Vex8tion:
> Sorry to say- I'm definitely not a fan of this PR- Batteries are Soteria's thing and should stay that way.

Lonestar can actually supply cells equivalent to Omega cells, except without a higher charge rate. These are two steps down from that; they don't conflict with SI's battery niche at all. There is an equivalent level of Power-Geyser, and a next higher tier of Power-Geyser, AND the Omega cells and self-recharging cells above these.

And Lonestar's Robustcell-X's, of course, though they don't charge as fast.

Vex8tion:
> Not a definitive no by any means, however, if it is decided that these clash with Soteria's battery-niche, I recommend making them craftable using high-end soteria cells, similar to how the handforged parts need interaction with soteria for tier 3 parts, these could require power-geysers.
> 
> I do love the sprites, they do fit in very well

I originally did consider making these craftable using existing battery cells. However, batteries on their own are by no means a trademark of any one faction (see my other comment: even Lonestar can produce Omega-level batteries, and these are far from that), and it makes sense that Guild, being intimately familiar with power generation and storage, would be able to craft batteries from scratch.

Moreover, the purpose of these is to fill the gap for when SI or LS isn't around to provide higher capacity batteries reliably. Omega and Robustcell-X batteries are both two steps higher in terms of max capacity, while Omegas are also twice again as fast to charge.

Vex8tion:
To break down the balance more quantitatively:
SI's premium non-self-recharging cells are the Omega line. These come in 1600M, 400S, and 20000L, all significantly higher capacity, and which charge at a max rate of 24% per tick, compared to the max rate of 12% per tick these batteries provide, or the 8% per tick that the Lonestar Robustcell-X line gives (which has the same capacity across the board as SI's Omegas).

The highest tier of Greyson Posi-Cells are equivalent in capacity to Omega cells, with the exception of L-size cells, which max out at the 16000L.

benj8560:
after hearing about it I was strongly meh but after a look...their hand crafted. Annoying to make. It's 20x easier to just throw mats at SI or arc welders at lonestar for cells that are generally better and cheaper. This is only going to be done out of necessity and is very far from a replacement for anyone. Or even a side grade.

benj8560:
(also note that guild cant even reach the max charge rate of these large hand crafted cells without a industrial recharger from research)

Vex8tion:
> (also note that guild cant even reach the max charge rate of the large hand crafted cells without a industrial recharger from research)

this explains why the big cell charged so damn slow in testing lol

ChefDoggo:
can't believe the slab is getting dunked on from this, surely this will fix the fact its absolute ass, surely

PR #5552, opened 2024-07-18
Update psi_catalysts.dm

benj8560:
Fixes the psi catalyst spawner for psi behemoths to correctly spawn Psionic Suture and Psionic Temperature Regulate at Trilbys request.

Dramatically buffs the spawn rates on both the behemoth and ponderous drop chance to make these new powers actually realistically obtainable.


benj8560:
also I will admit the drop chance _might_ be overtuned. I'll see how it is in practice and follow up with a reduction if that's the case

PR #5555, opened 2024-07-20
For surgery, shift it all down one tile

hex37:
NOTE: Only Merge after the Face-lift PR is merged

- Restructures everything to give surgery 5 tiles of space


benj8560:
thiiiink you could plug in a small fix for guild in here to real quick? Just set the smelter in the smelter rooms input_side = 8

benj8560:
they never changed its side vars when they added it...it defaults to a config that throws people down disposals ;-;

benj8560:
that or speedmerge so I can without conflicts _nod_

PR #5560, opened 2024-07-21
Hopefully fixes lodge bolt lance and minor smelter fixes.

benj8560:
requesting a TM since I'd like to play with these lodge numbers on live and tweak them more. Im still somewhat stuck. supereffectivness on projectile weapons is *still* not working and I don't know why. If anyone has ideas lemme know.

IMPORTANT NOTE: Merge #5561 before TMing this
Changelog:

Fixes the prebuilt guild smelter. It now has space OSHA compliant settings by default and hopefully should reduce the the number of adepts down disposals.
Fixes the desc if 9mm ball boxes. They  no longer declare to be non-lethal

Nerfs mukwah knocking them down to 300 health from 400. (way to chonky on new combat)
Nerfs slepnirs down to 350 from 550. (hypersonic speeds plus crazy health was making them pretty much guaranteed death)

Buffs lodge bolt lance projectiles. generally just higher damage.

### DIFF
--- a/code/game/machinery/smelter.dm
+++ b/code/game/machinery/smelter.dm
@@ -33,6 +33,8 @@
 
 /obj/machinery/smelter/cargo_t2_parts
 	input_side = WEST
+	output_side = NORTH
+	refuse_output_side = EAST
 
 /obj/machinery/smelter/cargo_t2_parts/Initialize()
 	. = ..()
--- a/code/game/machinery/smelter.dm
+++ b/code/game/machinery/smelter.dm
@@ -32,6 +32,7 @@
 	var/scrap_multiplier = 0.5 //50% refunds
 
 /obj/machinery/smelter/cargo_t2_parts
+	input_side = WEST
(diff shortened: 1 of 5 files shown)


Dissentbay:
By the way, the smelter is set in map like that too, so you might want to replace the current mapped one whenever you can

benj8560:
the smelter is map edited AND set incorrectly? AHHH

Dissentbay:
no, it's set the same way you did, but in map

benj8560:
> no, it's set the same way you did, but in map

it is quiet without a doubt. **not**

benj8560:
but I'll double check here

benj8560:
H3yvw6P7Bq


> no, it's set the same way you did, but in map


 freshly booted soj off master branch

Dissentbay:
WHAT THE
I tested it when I pr'ed it, it'was exactly like you coded in
image

PR #5570, opened 2024-07-23
Cloth psych ward walls

benj8560:
Was messing around and HOH figured out you can make cloth walls. perfect for are psych ward! I was crazy once.....

dreamseeker_75tkLLykUk

Changelog:

Adds cloth walls to the psych ward rooms

### DIFF
--- a/code/game/turfs/simulated/wall_types.dm
+++ b/code/game/turfs/simulated/wall_types.dm
@@ -230,6 +230,9 @@
 /turf/simulated/wall/wood/New(var/newloc)
 	..(newloc,MATERIAL_WOOD)
 
+/turf/simulated/wall/cloth/New(var/newloc)
+	..(newloc,MATERIAL_CLOTH)
+
 
 /turf/simulated/wall/marble/New(var/newloc)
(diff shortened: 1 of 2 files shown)


hex37:
On one hand I think that the psych ward may not be long for this world.

On the other, this is very non-intrusive and on-point. So if I'm not going to remove it any time soon, why not?

cdb-is-not-good:
Question; if these don't stop injury when you leap into them, what is the point?

benj8560:
> Question; if these don't stop injury when you leap into them, what is the point?

looks nice? Does there need to be a point. 



> On one hand I think that the psych ward may not be long for this world.
> 
> On the other, this is very non-intrusive and on-point. So if I'm not going to remove it any time soon, why not?

Why remove?

benj8560:
also I highly doubt someone _could_ go beyond midly hurting themselfs with diving. Once your at 40 or something brute it does the ''to hurt to dive'' thing. What you should be thinking about is someone smashing the light with their shoes and you knowwww

benj8560:
diving into walls does not hurt you at all as is. Stawp gaslighting me ;-;

PR #5682, opened 2024-09-18
Removal of a second chef and bartender slot

MikeTheWallDweller:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>
Very simple change, makes it so there can be only one chef, and only one bartender. You might ask - why? It is because every chef sees the kitchen as their own. It's a ruthless world out there, and there can be only one who cooks. And- Two bartenders is just overkill, honestly.

	
<hr>
</details>

## Changelog
:cl:
tweak: changed the amount of job slots for cooks and bartenders from 2 to 1

/:cl:

### DIFF
--- a/code/game/jobs/job/civilian.dm
+++ b/code/game/jobs/job/civilian.dm
@@ -4,8 +4,8 @@
 	department = DEPARTMENT_LSS
 	department_flag = LSS
 	faction = MAP_FACTION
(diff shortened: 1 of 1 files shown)


MikeTheWallDweller:
Who knows? Perhaps there could be a waiter job in the future, too... 👉👈

PR #5684, opened 2024-09-19
New job titles, fixing spelling errors and rephrasing some descriptions

Smokingsquirrel:
Adds new Job titles for:

Corpsman - Field Medic
Marshal officer - Marshal Patrolman, Marshal Enforcer

MD - Chemist added and line about there being a different chemist for the department removed
ELT - Soteria Biolab Enforcer and Emergency Recovery Specialist

### DIFF
--- a/code/game/jobs/job/medical.dm
+++ b/code/game/jobs/job/medical.dm
@@ -169,6 +169,7 @@
 	supervisors = "the Chief Biolab Overseer"
 	difficulty = "Ungratifying."
 	selection_color = "#a8b69a"
+	alt_titles = list("Soteria Biolab Enforcer", "Emergency Recovery Specialist")
 	wage = WAGE_PROFESSIONAL
 	outfit_type = /decl/hierarchy/outfit/job/medical/recovery_team
 	disallow_species = list(FORM_AGSYNTH, FORM_CHURCHSYNTH, FORM_NASHEF)
--- a/code/game/jobs/job/medical.dm
+++ b/code/game/jobs/job/medical.dm
@@ -72,7 +72,7 @@
 	selection_color = "#a8b69a"
 	wage = WAGE_PROFESSIONAL
 	minimum_character_age = 25
-	alt_titles = list("Soteria Nurse"=/decl/hierarchy/outfit/job/medical/doctor/medNurse, "Soteria Emergency Physician"=/decl/hierarchy/outfit/job/medical/doctor/medERPhys, "Soteria Surgeon"=/decl/hierarchy/outfit/job/medical/doctor/medSurgeon, "Soteria Chemist")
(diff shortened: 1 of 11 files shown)


Smokingsquirrel:
Got no idea why the corrections on security and medical are going through as failed, if anyone has an idea pray tell.

code/game/jobs/job/medical.dm
```diff
@@ -94,15 +94,13 @@
...
-	Due to the nature of your work, you may find yourself confined to the department for the shift majority. Don't abandon patients."
+	description = "The Doctor is a professional medic and surgeon dedicated to healing the sick and injured at all costs.<br>\
+	A broad range of medical procedures fall under your purview - diagnostics, general treatment, surgery and virology.<br>\
+	You are not expected to be an expert in all: specializing in an area is fine. Divide tasks amongst colleagues with CBO guidance.<br>\
+	Due to the nature of your work you may find yourself confined to the department for the shift majority. Don't abandon patients."
 
 	duties = "Heal the sick and injured, whatever their complaint.<br>\
-		Diagnose illnesses, offer general services, perform surgery, or even study viruses.<br>\
-		Fill in at chemistry if a Chemist is unavailable."
+		Diagnose illnesses, offer general services, perform surgery, or even study viruses.<br>\"
```
BurpleBineapple, line 103: Checks are failing because of the way you've edited the blurbs. You need to remove this <br>\. Anything that follows a \ string is ignored until the next line. This is normally used to organize text to fit on screen without having to scroll as a code cleanliness thing.

There might be others but this is the one that stood out. I didn't review all of this.

Smokingsquirrel: Ahh so it shouldn’t end on that, alright

cdb-is-not-good:
All fine except for QM for spec. Ideally jobs with only one slot shouldn't have alt titles- and definitely not alt titles that are jobs on other servers that could confuse their purpose.

Smokingsquirrel:
> I wouldn't expect medical treatment from a chemist.

Thing is that the former chemist is now the ELT so unless someone codes a completely new chemist role, MD‘s already do that anyway. I don’t expect a NURSE to do chemical stuff either but that’s a problem the server made when omitting the chemist.

du-top:
A nurse pushing buttons according to a receipt makes more sense than that someone who only learned pushing buttons grabs a scalpel and performs surgery. But yeah, I'm biased, so are you (I mean, you did it like that for a reason), let's see what others think.

Smokingsquirrel:
> A nurse pushing buttons according to a receipt makes more sense than that someone who only learned pushing buttons grabs a scalpel and performs surgery. But yeah, I'm biased, so are you (I mean, you did it like that for a reason), let's see what others think.

I made this because the MD titles already had the perfect grounds for that. I also thought about adding it to ELT at first but the description still states they arent really proper medical personnel so i wouldnt trust those to mix chemicals either. Its easy ingame but it really should be someone who properly knows what they are doing, same goes for the fact that not every MD should be a know it all - but is due to gameplay reasons.

cdb-is-not-good:
Thumbs up one hundred one hundred(concept wise I haven't looked at the code but I severely doubt there's any errors in something like this)

PR #5710, opened 2024-09-28
Gun parts fixes

Rat-Time:
Adds wooden grip to the list of accepted grips on the greasegun frame and texan to the resultvars so texan SMGs can be disassembled and reassembled properly.
Changes the bakelite grip on sawn down STS rifles to wood in order to match the current version of the full rifle.


cdb-is-not-good:
....No. The texan is a higher end SMG and not every gun needs to be on that stupid system.

cdb-is-not-good:
Why was this fullmerged after less than a day

PR #5712, opened 2024-09-29
Allow items in storage slots to not hassle body cover checks

Ryan180602:
## About The Pull Request
<details>
<summary>
	Fixes the bug about head covers in storage slots stopping implanters. though i don't think this is the best solution, because headwear (and some other gear) may be able to be placed in empty belt slots as well, which can prove an issue. would be good to overhaul how body-part checks are done in the future
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
fix: Storage slots don't strike bodypart cover checks
/:cl:


king5327:
Solves an actual problem, very good pr, all that jazz.

PR #5717, opened 2024-10-02
Swamp revamp

GlitchySoftware:
Changed Southernmost side of the Swamp, which used to have some caves into a full-blown site for exploration and looting. 
THIS WILL NEED REBALANCING IN TERMS OF LOOT/MOBS
image


Bamhalazam:
beast cave more like gooncave lmao

srsly tho, good shit

PR #5720, opened 2024-10-03
Roach Meat Audit + GP station change (And Misc Station tweaks)

WilsonWeave:
Meat and GP

# The big change
Alright, so first and foremost the Dayanji station (Secondary Greyson Station) has had it's tool/part stock replaced with the special Greyson Zhengdou voidsuit (priced at 25k) and the GP Greyson air tank (priced at 2k)
This is because I personally believe it allows LS to stomp on what should be the Guild's and Soteria's toes way too easily, given it's just a matter of nabbing some plat from the mines and using the SOM phone to immediately phone in the station. I'm planning on tossing in more misc. Greyson goodies in the near future, but this is a decent start.

## The less significant changes
The roach meat offers have been audited, generally a 1/3 to 1/2 reduction in value depending on the specific meat. Given Armitage allows players to buy roach cubes obscenely cheaply, this was long overdue. Also fixes jager meat being listed as a repeat seuche offer.

Cad has been given an offer for generic spider meat, under the presumption that antidotes can be procured from the toxins in the meat. 

Recoll has been given an offer for termite meat. They can probably sift out some more ore from the meat.

Together, the last two changes gives more incentive to butcher the total variety of maint bugs to pawn off, rather than just leaving roach meat as the cash king option. Though, I didn't split the meats off into their sub-types because it's annoying having to parse the meat types out so the parent meat-type doesn't gobble up all of the categories of meat at a reduced price.

Armitage has had the Kaiseraurum bottle commented out, because kaiseraurum doesn't exist and you'd just get an empty bottle for a thousand credits. Also, bumps up the price on the offer for zombie powder.

Garbaj has had it's default value of 32k favor (yikes!) reeled back to 8k, also very long overdue.

### Proof that this all works:

image
_They can in fact be imported, and the voidsuit does in fact come with a helmet_

image
_Still plenty lucrative!_

image
_Cad was hurting for more means of being favored tbh_

image
_Recoll wasn't hurting for more means of being favored, but I don't think anyone would complain_

image
_No more being scammed into buying an empty bottle for a thousand credits_

image
_No more bartenders toiling away to make drinks to settle 1.5+ hours of favors to get a single station's  hidden inventory!_

TLDR/Changelog friendly consolidation

- LS can't order GP Tool and stockpart disks anymore
- LS can now order the Zhengdou voidsuit and oxygen generator tank at gouged prices
- Roach meat prices have been audited and cut down, however spider and termite meat are now offers for cad and recoll. Also fixes an error in the roach meat listings.
- Armitage pays a bit more for zombie powder bottles now, can't buy kaiseraurum anymore because it doesn't exist
- Garbaj favor/hidden inventory favor has been significantly reduced to 8k from 32k

### DIFF
--- a/code/modules/trade/datums/trade_stations_presets/1-common/caduceus.dm
+++ b/code/modules/trade/datums/trade_stations_presets/1-common/caduceus.dm
@@ -122,6 +122,7 @@
 		/datum/reagent/medicine/bicaridine = offer_data("bicard bottle (60u)", 250, 3),
 		/datum/reagent/medicine/kelotane = offer_data("kelotane bottle (60u)", 250, 3),
 		/datum/reagent/medicine/dylovene = offer_data("dylovene bottle (60u)", 250, 3),
+		/obj/item/reagent_containers/food/snacks/meat/spider = offer_data("spider meat", 250, 5),
 		/datum/reagent/nanites/uncapped/control_booster_utility = offer_data("Control Booster Utility bottle (60u)", 30000, 1),
 		/datum/reagent/nanites/uncapped/control_booster_combat = offer_data("Control Booster Combat bottle (60u)", 30000, 1)
 		)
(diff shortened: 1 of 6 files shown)


Smokingsquirrel:
As a frequent LS Player, I dont have anything bad to say about the balancing. The few new additions for favoring are also doable and allow less experienced LS Players to also work towards the favoring-goal.

Greyson Tools are already easily available with enough SI, Prospie or Guildie players.

PR #5753, opened 2024-10-12
Clean up poison attacks/give wasprender poisonbite

Ryan180602:
## About The Pull Request
<details>
<summary>
	Instead of having multiple inherited overrides of `UnarmedAttack`, standardises it for both `superior_animal` and `hostile`. Also gives the new wasp render the same poison bite as wasps (except with double the poison)

Now any mob under `hostile` or `superior_animal` can have poison injection added tidily
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:
balance: Wasp renders also inject upto 6u wasp inflammation toxin on a successful, unblocked hit.
/:cl:


Ryan180602:
does not fix the issue of renders not attacking turrets. but is tested and should still work as it always does, now with the added benefit of having poison damage for those that need it

PR #5756, opened 2024-10-13
Solarian armor added helmet and mask for full set no kit bashing

Karlov357:
## About The Pull Request
<details>
<summary>
addition of a helmet and gas mask for the chemical corps Solarian armor so there is a full set of armor that matches or looks better than mixing and matching different helmets to fit the vibe of the armor. Mask is a general gasmask but with 100 rad protection, helmet is essentially the ballistic mask from sec disk minus one point and some added rad and bio protection to fit the armors intention. Both mask and helmet are added to arau favored inventory next to the Solarian armor
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl:

/:cl:


code/modules/clothing/head/helmet.dm
```diff
@@ -257,6 +257,58 @@
 		usr.update_action_buttons()
 		return 1
 
+/obj/item/clothing/head/helmet/hazcap
+	name = "CBRN Helmet"
```
Lamasmaster, line 261: CBRN helmet - lowercase so it's same as the other helmets

Karlov357: fixed

PR #5774, opened 2024-10-20
Soteria Buff one out of one hundred.

Legate-Maxson:
It's been a long time coming, This PR was meant to be done months ago. It reduces the research points needed by 37500 because it removes the research tree for the unmaker. The Greyson portable shield was moved to the glass widow branch. Unmakers can still be purchased from cargo, Found at the end of the gauntlet and now are a rare drop from the Greyson Titan robots that can be found at the end of the gauntlet or the boss rooms in the Greyson's outpost.

The node was left untouched as things will be added too it later for a replacement, but spiriting needs to be done. Greyson oxygen tank will most likely added to this tree as well in the future.


Stutternov:
truly the buff soteria always needed...........

PR #5837, opened 2024-11-12
Lodge Map Rework

Rat-Time:
Lodge map rework.  No more underground, all interesting cave features have been migrated to the surface along with materials and rewards useful to lodge hunters.
The endless xenos were removed, as was the greyson facility.  Sargoyles and tribal birds have an increased presence and their own locations, as well as lairs for the solitary large tengolo.  

StrongDMM_uoTMMkbStS


RagingLemons1:
RIP lodge greysons, I never got to shoot a hunter asking why we were raiding (((their))) dungeon

Rat-Time:
Added a greyson facility, moved the renders into the xeno cave and the xenos into the larger render cave.
StrongDMM_XzkHZospKr
StrongDMM_od4C69Nd1d
StrongDMM_mxdFFihYYL

PR #5867, opened 2024-11-23
Adjusts destructive analyzer so that stock part changes actually matter (Unfucked and Reloaded)

wshuwshuwshi:
## About The Pull Request
Removes the 0.1 multiplier from the stock part calculations in the destructive analyzer so that upgrading stock parts actually does something. Please test merge this so we can feel out how it is balancewise
<summary>

	
<hr>
</details>

## Changelog
:cl:
tweak: Removed the 0.1 multiplier from stock part calculations
/:cl:

### DIFF
--- a/code/modules/research/destructive_analyzer.dm
+++ b/code/modules/research/destructive_analyzer.dm
@@ -28,7 +28,7 @@ Note: Must be placed within 3 tiles of the R&D Console
 	var/T = 0
 	for(var/obj/item/stock_parts/S in src)
 		T += S.rating
-	decon_mod = T * 0.1
+	decon_mod = T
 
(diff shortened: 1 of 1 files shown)


Trilbyspaceclone:
How this works:
lets say stockpart is level 3
3 x 0.1 = 0.3 aka a return of 30% of lost material
lets say now with your new change I have a level 3 stockpart
3x1 = 3 meaning I get ***200%*** materials aka making a materal doup

Trilbyspaceclone:
image
Tested on a local, its a mater doup.
I also see theirs no decon cap so level 4 parts (guildfab/exl) also do this behavior

AMcilraith:
My personal suggestion would simply be as follows: 

Add a 1.N modifier, with N being the tier of the part.

PR #5881, opened 2024-11-28
Adds a church robe

Smokingsquirrel:
Adds a church robe and replaces the old white church towel robe.

Style is added to Church and Vector Vestments.

image


code/modules/clothing/under/jobs/civilian.dm
```diff
@@ -1,4 +1,4 @@
-//Alphabetical order of civilian jobs.
+f//Alphabetical order of civilian jobs.
```
Smokingsquirrel, line 1:
```suggestion
//Alphabetical order of civilian jobs.
```

PR #5900, opened 2024-12-07
Xenobio batong

Lamasmaster:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>
	Fixes a spelling misstake
<hr>
</details>

## Changelog
:cl:
fix: Typo in xenobio stun baton description
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/tools/stunbatons.dm
+++ b/code/game/objects/items/weapons/tools/stunbatons.dm
@@ -253,7 +253,7 @@
 	force = WEAPON_FORCE_NORMAL
 	throwforce = WEAPON_FORCE_NORMAL
 	stunforce = 0
-	agonyforce = 30
+	agonyforce = 50
 	hitcost = 125
 	attack_verb = list("poked")
 	slime_killer = TRUE
--- a/code/game/objects/items/weapons/tools/stunbatons.dm
+++ b/code/game/objects/items/weapons/tools/stunbatons.dm
@@ -253,7 +253,7 @@
 	force = WEAPON_FORCE_NORMAL
 	throwforce = WEAPON_FORCE_NORMAL
 	stunforce = 0
-	agonyforce = 20 //No
+	agonyforce = 30
 	hitcost = 125
 	attack_verb = list("poked")
 	slime_killer = TRUE
--- a/code/game/objects/items/weapons/tools/stunbatons.dm
+++ b/code/game/objects/items/weapons/tools/stunbatons.dm
@@ -247,13 +247,13 @@
 
 /obj/item/tool/baton/slimebaton
 	name = "xenobio baton"
-	desc = "A stunbaton that is designed against slimes and other lab misstakes."
+	desc = "A stunbaton that is designed against slimes and other lab mistakes."
 	icon_state = "prod_si"
 	item_state = "prod_si"
 	force = WEAPON_FORCE_NORMAL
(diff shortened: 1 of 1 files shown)


du-top:
Salt PR because addi loves that baton very verbally?

Lamasmaster:
No because some things are long overdue.

Smokingsquirrel:
So the old one had 50 agony and the security stun baton along with the improvised one has 40 agony. The Gloves have 30. 
I agree on 50 being way too much but 20 is probably not enough. 

I would suggest putting 25.

code/game/objects/items/weapons/tools/stunbatons.dm
```diff
@@ -247,13 +247,13 @@
...
 	name = "xenobio baton"
-	desc = "A stunbaton that is designed against slimes and other lab misstakes."
+	desc = "A stunbaton that is designed against slimes and other lab mistakes."
 	icon_state = "prod_si"
 	item_state = "prod_si"
 	force = WEAPON_FORCE_NORMAL
 	throwforce = WEAPON_FORCE_NORMAL
 	stunforce = 0
-	agonyforce = 50
+	agonyforce = 20 //No
```
Trilbyspaceclone, line 256:
```suggestion
	agonyforce = 30
```
Its expensive, and you can do the same meme setups to 1 tap everyone (and better) with normal baton you can print with a sec disk/genetics free baton

The issue is that tool upgrades for hit cost directly also mult agonyforce as seen [here](https://github.com/sojourn-13/sojourn-station/blob/master/code/game/objects/items/weapons/tools/mods/_upgrades.dm#L371).

Trilbyspaceclone (changes requested):

ChefDoggo:
image

AMcilraith:
I'm going to cite both the contributor guidelines, and say: You cannot nerf it more than 25. That's over the 50% nerf we permit.

Secondly, it is extremely expensive to get it to the quantity we do. At least 100k research points go into getting the greyson parts unlocked, platinum, gold, silver.

It's very much a late-round gain. 50 is also not too much when you consider how slimes are handled. The one-tap baton is HEAVILY upgraded with HIGHLY EXPENSIVE parts. It's not simple.

Also, for the record, it does NOT one tap people. One tapping implies it takes one hit and it's done. It stuns them, and at around minimum robustness, even with one hit, it takes upwards of 30-40 seconds for the person to go down.

Compare that with a more robust person, i've seen some take 3-4 hits before that thirty second mark even becomes an issue.

This is clearly a salt-PR. I'm advising it be denied.

AMcilraith:
Also dropping it from 50 to 25/30 will prevent it from killing slimes in one hit.

That needs to be fixed in some way.,

Smokingsquirrel:
Okay given the explanation and costs included, 20 does seem a bit low lmao.

Trilbyspaceclone (commented):
I still need to think this over more.
Its major use is 1 tapping slimes, with a newer side use of anti-hostile mobs in genetics I guess?
Like its more expensive, cost more per hit and if its less good then a normal baton then why get it (outside of xenoarch).

Trilbyspaceclone:
After a long thing and such. I do not think this is a good change.
It does not properly fix the issue of batons being some of the best scaling weapons in the game.

A better fix would be to change how much agony force force is scaled up by power cost as stated in previous reviews.

The spelling fix is approved tho.

PR #6011, opened 2025-04-29
Yet Another Cargo Tweaking (I'm tweaking)

WilsonWeave:
Removes plain meat from being sold from most stations.
Spider and Many variants of roach meat are no longer offers on stations, their chems in bottles are now what are sought after.
Lowers the value of the Bicaridine and Kelotane offers, adds plenty of more advanced medicine offers to be sought out by competent chemists.
Buffs the offers of many Guild crafted components and weapons in Hellcat.
Assigns prices to the animal crates listings in Faith.
Faith no longer wants slabs of meat, they want cutlets instead.
Slightly nerfed Dionis meat offer, gave them roach chem offers (Flavoring for Cht'Mant clientele, or something.)
Armitage chems bumped up in price. They're still lower than the offers, so they could stand to be bumped up even more in the future if this proves problematic.

image
image


WilsonWeave:
Reducing redundant/duplicate offers from offers is a worthwhile pursuit in reducing the headaches cargo has to deal with, instead of trying to hash out how much they owe somebody after having plantbags full of meat dumped at their feet. They instead can either use a kitchen grinder, ask SI for a grinder, or gamble with circuit board disks to source their own grinder and just see how many bottles of different reagents they get to export and approximate a price from there.

PR #6074, opened 2025-09-22
Crafting station doup fix

Trilbyspaceclone:
## About The Pull Request
Makes crafting stations no longer doup materals


Trilbyspaceclone:
Resolves in ##6046

